Issue #19, opened 2026-04-27
fix(scripts): override iam.allowedPolicyMemberDomains in provisioner; correct doc command


va-worker:
## Problem Statement

Two related defects discovered live during workshop pre-provisioning on 2026-04-27:

1. **The `iam.allowedPolicyMemberDomains` org constraint blocks every participant IAM binding** when the binding target is an external-domain identity (personal Gmail, non-allowlisted Workspace). The current `provision-gcp-project.sh` does not handle this — it provisions the project + SA + APIs, then `provision-workshop-roster.sh` calls `gcloud projects add-iam-policy-binding --member="user:$email"` and `gcloud` rejects with `FAILED_PRECONDITION`. There is no retry or workaround in either script. The facilitator must run a manual `gcloud resource-manager org-policies set-policy` per project before the wrapper succeeds.

2. **The two doc PRs that captured this gotcha (`agile-flow-gcp`#18 pattern #30 and `agile-flow-meta`#87 runbook §5) prescribe the wrong gcloud command.** Both use `disable-enforce`, which only works on **boolean** constraints. `iam.allowedPolicyMemberDomains` is a **list** constraint (with `allowedValues` / `allValues`), so `disable-enforce` returns:

   ```text
   ERROR: (gcloud.resource-manager.org-policies.disable-enforce) INVALID_ARGUMENT:
   Policy and Constraint must be of the same type
   ```

   The correct command is `set-policy` with a YAML/JSON body of `listPolicy: { allValues: ALLOW }`. Confirmed working live.

This is the third propagation-class issue the workshop has hit (after #14 API-enable 403, #16 SA-create 404). Pattern: GCP defaults aren't built for short-lived multi-tenant projects, and the script must absorb the workarounds.

**Parent Epic:** vibeacademy/agile-flow-gcp#2 — Workshop-ready hardening of the GCP fork
**Effort Estimate:** S
**Priority:** P0 (dry-run is 2026-05-01, 4 days from filing; current behavior requires manual intervention per participant project)

---

## A. Environment Context

- Stack: Bash, same as #14 / #16
- Files to modify:
  - MODIFY `scripts/provision-gcp-project.sh` — add a "step 4.5: override domain-restricted-sharing" block, gated on the constraint being detected as enforced
  - MODIFY `scripts/provision-gcp-project.test.sh` — add a scenario covering the override (`gcloud` stub returns enforced, override sets the project policy)
[...]

Issue #6, opened 2026-04-17
feat(infra): budget cap + auto-cutoff Cloud Function for participant projects


tck517:
## Problem Statement

A GCP workshop has no spending guardrails today. Each participant project needs visible budget alerts so the facilitator gets notified before runaway spend turns into a real bill. Auto-cutoff via Cloud Function is deferred to a post-workshop ticket — for ≤8 participants over 2 days, a $25 alert per project (max blast radius ~$200) is sufficient when paired with facilitator-side billing monitoring.

**Parent Epic:** vibeacademy/agile-flow-gcp#2 — Workshop-ready hardening of the GCP fork
**Effort Estimate:** S
**Priority:** P0 (workshop-prep, last dry-run blocker)

---

## A. Environment Context

- Stack: Bash + `gcloud billing budgets create` (API: `billingbudgets.googleapis.com`)
- Notifications: rely on the budget's default IAM recipients (Billing Admins/Users on the billing account). No separate Cloud Monitoring notification channels — fewer moving parts, no extra API to enable.
- Files to modify:
  - MODIFY `scripts/provision-gcp-project.sh` — add Step 5.6 between Step 5.5 (WIF) and Step 5.7 (Neon)
  - MODIFY `scripts/provision-gcp-project.test.sh` — stub-test the budget step
  - MODIFY `docs/PLATFORM-GUIDE.md` — document the alert behaviour
- Existing pattern: Step 5.5 / Step 5.7 — gated on env vars, idempotent (describe-or-create), test-stubbed via PATH-injected `gcloud`

---

## B. Guardrails

- Cap parameterized via env var `BUDGET_CAP_USD`. When unset → skip silently (non-workshop usage).
- Step is idempotent: re-running on a project that already has the budget should not create duplicates. Lookup by display-name `af-workshop-cap-<project_id>`.
- Threshold rules: 50%, 90%, 100% (current spend basis) and 100% forecasted.
- Budget MUST be scoped to the single participant project via `--filter-projects=projects/<PROJECT_NUMBER>`, not the whole billing account.
- Required API: `billingbudgets.googleapis.com` — added to Step 2's enable list.
- The facilitator (script runner) needs `roles/billing.costsManager` on the billing account. Documented but not granted from inside the script.
- Auto-cutoff (Cloud Function) is explicitly out of scope; tracked in #42.

---

## C. Happy Path

[...]

Issue #5, opened 2026-04-17
feat(scripts): automate Workload Identity Federation in provision-gcp-project.sh


tck517:
## Problem Statement

`scripts/provision-gcp-project.sh` only automates the service-account-key shortcut (`--with-sa-key`). For workshops, every participant project needs Workload Identity Federation set up so their GitHub repo can deploy to Cloud Run without a long-lived JSON key. Today this is a manual `docs/PLATFORM-GUIDE.md` Step 5 walkthrough — too slow at workshop scale and prone to mistakes.

**Parent Epic:** vibeacademy/agile-flow-gcp#2 — Workshop-ready hardening of the GCP fork
**Effort Estimate:** M
**Priority:** P1

---

## A. Environment Context

- Stack: Bash + `gcloud iam workload-identity-pools` commands
- Reference: Google's "Configuring Workload Identity Federation" docs for the exact command sequence (pool create, provider create, IAM binding)
- Files to create or modify:
  - MODIFY `scripts/provision-gcp-project.sh` — add `--with-wif GITHUB_REPO` flag that creates the pool, provider, and binding for `repo:OWNER/REPO`
  - MODIFY `docs/PLATFORM-GUIDE.md` — replace the manual Step 5 with the new flag invocation
  - MODIFY `scripts/provision-workshop-roster.sh` (created in A2) — wire `--with-wif` per row using `roster.csv`'s `github_user/repo` field
- Existing workflow files (`deploy.yml`, `preview-deploy.yml`) already consume `GCP_WORKLOAD_IDENTITY_PROVIDER` and `GCP_SERVICE_ACCOUNT` secrets — no workflow changes needed

---

## B. Guardrails

- Do NOT remove the `--with-sa-key` flag — keep it as a fallback for participants who hit org-policy blocks on WIF.
- Do NOT create the WIF pool more than once per project (idempotent — check `gcloud iam workload-identity-pools describe` first).
- The IAM binding MUST be scoped to `repo:OWNER/REPO` (not the whole GitHub org). Workshop participants have one repo each.
- Output MUST print the exact `GCP_WORKLOAD_IDENTITY_PROVIDER` and `GCP_SERVICE_ACCOUNT` values for the participant to paste into GitHub repo secrets.
- Do NOT print or log the SA key contents anywhere when WIF is in use.

---

## C. Happy Path

1. Facilitator runs:
   ```
   GCP_PROJECT_ID=af-alice-2026-05 \
[...]

PR #21, opened 2026-04-27
feat(scripts): workshop-setup.sh + workshop-teardown.sh wrappers

### LINKED ISSUE #20: feat(scripts): workshop-setup.sh + workshop-teardown.sh wrappers
## Problem Statement

Workshop facilitators currently run a sequence of commands to bring up and tear down a workshop:

```text
1. (precondition) gcloud auth login as the facilitator
2. (precondition) Verify billing account access
3. (precondition) Verify gcloud is authenticated to the right project default
4. ./scripts/provision-workshop-roster.sh roster.csv
5. (manual) gcloud resource-manager org-policies set-policy ... per project
   (until #19 lands)
6. (post-workshop) loop gcloud projects delete af-* --quiet
```

Steps 1–3 fail silently when something is off — the facilitator only finds out at step 4 when an inscrutable gcloud error fires mid-loop. Step 6 is "the gcloud loop you remember" with no canonical home in the repo, so each cohort the facilitator pastes the loop fresh.

This ticket adds two thin wrapper scripts so the facilitator's mental model is two commands: setup, teardown.

**Effort Estimate:** S
**Priority:** P1 (workshop is 9 days out; no manual workaround burden today, but every step that's automatable is one less failure mode at dry-run)

---

## A. Environment Context

- Stack: Bash, same as #14 / #16 / #19
- Files to create:
  - CREATE `scripts/workshop-setup.sh` — pre-flight checks + delegate to `provision-workshop-roster.sh`
  - CREATE `scripts/workshop-teardown.sh` — confirm-then-loop `gcloud projects delete` over the roster
  - CREATE `scripts/workshop-setup.test.sh` — mocked-shell tests for the pre-flight logic
  - CREATE `scripts/workshop-teardown.test.sh` — mocked-shell tests for the deletion loop
- Files to modify:
  - MODIFY `docs/PLATFORM-GUIDE.md` — replace the "Workshop: Provisioning N Projects" section with a setup-then-teardown lifecycle, pointing the canonical entry points to the new wrappers
- Reference:
  - `scripts/provision-workshop-roster.sh` — existing N-project wrapper (called by setup)
  - `scripts/provision-gcp-project.sh` — existing single-project script (delegated to by the above)

---

## B. Guardrails

- **Thin wrappers, not new logic.** `workshop-setup.sh` calls `provision-workshop-roster.sh` after pre-flight; it does NOT re-implement provisioning. `workshop-teardown.sh` only deletes projects; it does NOT touch billing accounts, folders, or org-level state.
- The setup pre-flight MUST check, in this order:
  1. `gcloud auth list` shows an active account (script exits 2 with a clear message if not)
  2. `gcloud billing accounts list` returns at least one row matching `BILLING_ACCOUNT_ID` and `OPEN=True`
  3. The roster CSV header matches `provision-workshop-roster.sh`'s expected header (`handle,github_user,email,cohort`)
  4. The roster has at least one data row
- The teardown MUST require explicit confirmation. Default behavior: prompt `[y/N]` with the list of project IDs that will be deleted. Override with `--yes` flag for non-interactive use (CI smoke tests, scripted teardown).
- The teardown MUST be idempotent: deleting an already-deleted project should log `[skip]` and continue, not error. Use `gcloud projects describe ... --format='value(lifecycleState)'` to check before deleting.
- Both scripts MUST be `set -euo pipefail` and pass `bash -n` and `shellcheck`.
- Both scripts MUST emit a clear summary at the end (counts: checked / created / deleted / skipped / failed) so a facilitator can paste the output into a session journal.
- Do NOT delete `roster.csv` (the input). Do delete `roster-output.csv` (the generated output) on teardown — it's stale once projects are gone.
- The teardown script MUST NOT delete projects whose ID does NOT match the `af-{handle}-{cohort}` pattern derived from the roster. This guards against a malformed roster taking down unrelated projects.

---

### Setup

```bash
BILLING_ACCOUNT_ID=XXX-XXX-XXX ./scripts/workshop-setup.sh roster.csv
```

1. Pre-flight prints `[ok]` for each check, `[fail]` and exit 2 if anything is wrong.
2. Hands off to `provision-workshop-roster.sh roster.csv`.
3. On success, prints a summary and points to `roster-output.csv`.

### Teardown

```bash
./scripts/workshop-teardown.sh roster.csv
# prints the list, asks "Delete N projects? [y/N]"
# user types y → loop runs

# OR for non-interactive use:
./scripts/workshop-teardown.sh roster.csv --yes
```

For each row, the wrapper:

1. Computes the project ID from the same `af-{handle}-{cohort}` pattern.
2. If the project is `ACTIVE`, calls `gcloud projects delete --quiet`.
3. If already `DELETE_REQUESTED` or absent, logs `[skip]` and continues.
4. Records each row's outcome (deleted / skipped / failed) in the summary.
5. After the loop, removes `roster-output.csv` if it exists.

---

## D. Definition of Done

- [ ] `scripts/workshop-setup.sh` exists, is executable, passes `bash -n` and `shellcheck`
- [ ] `scripts/workshop-teardown.sh` exists, is executable, passes `bash -n` and `shellcheck`
- [ ] Pre-flight checks fire in the order listed in section B and produce actionable error messages on failure
- [ ] Teardown prompts for confirmation by default; `--yes` skips the prompt
- [ ] Teardown is idempotent — re-running on an already-torn-down classroom logs all rows as `[skip]` and exits 0
- [ ] Teardown only operates on `af-{handle}-{cohort}` IDs derived from the roster — never against arbitrary project IDs
- [ ] `scripts/workshop-setup.test.sh` covers: missing `gcloud auth`, missing `BILLING_ACCOUNT_ID`, missing roster file, bad header, empty roster
- [ ] `scripts/workshop-teardown.test.sh` covers: happy path (all rows ACTIVE), idempotent re-run (all rows DELETE_REQUESTED), `--yes` skips prompt, malformed CSV is rejected
- [ ] All existing tests still pass (no regression on `provision-workshop-roster.test.sh` 14/14 or `provision-gcp-project.test.sh` 14/14)
- [ ] `docs/PLATFORM-GUIDE.md` "Workshop: Provisioning N Projects" section updates to point at the new entry points; the underlying `provision-workshop-roster.sh` is still documented as the implementation but no longer the recommended caller
- [ ] Real-GCP smoke deferred to the 2026-05-01 dry-run

---

## Out of scope

- Org-policy override automation — separate ticket #19, lands in `provision-gcp-project.sh`. Setup wrapper does not duplicate.
- Sending participant notification emails — runbook responsibility, not scriptable in a generic way.
- Exporting workshop session journal — manual, post-workshop.
- Folder creation for the workshop (T-1 week step) — facilitator does this once per cohort; not worth scripting.

va-worker:
## Summary

Closes #20.

Two thin wrappers giving facilitators a clean lifecycle:

```bash
BILLING_ACCOUNT_ID=… ./scripts/workshop-setup.sh    roster.csv
./scripts/workshop-teardown.sh roster.csv [--yes]
```

## What's in the PR

- **`scripts/workshop-setup.sh`** (~115 lines) — four pre-flight checks (gcloud auth, billing OPEN, roster header, data rows present), then `exec`s `provision-workshop-roster.sh`. No provisioning logic duplicated.
- **`scripts/workshop-teardown.sh`** (~125 lines) — confirms-then-deletes; idempotent on re-runs; only touches IDs derived from the roster (scope guard).
- **`scripts/workshop-setup.test.sh`** — 17 assertions, 6 scenarios.
- **`scripts/workshop-teardown.test.sh`** — 14 assertions, 6 scenarios.
- **`docs/PLATFORM-GUIDE.md`** — replaces the "Workshop: Provisioning N Projects" section with a setup/teardown lifecycle that points at the new entry points. Underlying scripts still documented as the implementation.

## Out of scope (intentional, called out in the doc)

- Org-policy override → #19
- WIF setup → #5
- Budget caps → #6
- Notification emails → meta runbook

## Test plan

- [x] `bash -n` and `shellcheck` clean on all four new scripts
- [x] `./scripts/workshop-setup.test.sh` — 17/17 pass
- [x] `./scripts/workshop-teardown.test.sh` — 14/14 pass
- [x] No regression: `provision-workshop-roster.test.sh` still 14/14, `provision-gcp-project.test.sh` still 14/14
- [x] markdownlint clean
- [ ] Real-GCP smoke at the 2026-05-01 dry-run

## Notes

- **Setup wrapper is `exec`-based** (line 116). It hands off to the inner wrapper without keeping a parent shell around, so the facilitator's terminal sees the inner script's output directly. Side effect: anything after the `exec` line is dead code (intended).
- **Teardown's scope guard.** Project IDs come from `awk`-parsing the roster's `handle` and `cohort` columns. The script never deletes an ID that wasn't derived from this roster, even if `gcloud` would otherwise allow it. Test 5 verifies this.
- **`--yes` is conservative.** Default is interactive; you have to opt in to non-interactive deletion. Test 3 verifies it doesn't block on closed stdin.

### DIFF
--- a/docs/PLATFORM-GUIDE.md
+++ b/docs/PLATFORM-GUIDE.md
@@ -227,12 +227,25 @@ If the workflow succeeds but the container fails its health check with
 
 ---
 
-## Workshop: Provisioning N Projects
+## Workshop: Lifecycle (Setup and Teardown)
 
-When running a workshop, the facilitator provisions one GCP project per
-participant. `scripts/provision-workshop-roster.sh` wraps
-`provision-gcp-project.sh` in a CSV-driven loop so all participant
-projects can be created with one command.
+When running a workshop, the facilitator's mental model is two commands:
+
+```bash
+# Bring up the classroom
+BILLING_ACCOUNT_ID=XXXXXX-XXXXXX-XXXXXX \
+  ./scripts/workshop-setup.sh roster.csv
+
+# Tear it down at T+1 day
+./scripts/workshop-teardown.sh roster.csv          # interactive prompt
+./scripts/workshop-teardown.sh roster.csv --yes    # non-interactive
+```
+
+The setup script runs four pre-flight checks (gcloud auth, billing
+account is OPEN, roster file exists with the expected header, roster
+has data rows) and then hands off to the underlying provisioning logic.
+Pre-flight failures exit 2 with actionable messages — far better than
+discovering a missing auth token mid-loop.
 
 ### Roster format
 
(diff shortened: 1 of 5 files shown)


va-reviewer:
## PR Review — #21

### Requirements

- [x] Both scripts exist, executable, shellcheck clean
- [x] Pre-flight checks fire in the documented order (gcloud auth → BILLING_ACCOUNT_ID → roster file/header → data rows present)
- [x] Teardown prompts by default; `--yes` skips
- [x] Teardown is idempotent (test 2 verifies via stub returning DELETE_REQUESTED)
- [x] Teardown only operates on roster-derived IDs (test 5 verifies via whitelist filter)
- [x] No regression — CI green; existing 28 assertions still pass
- [x] `docs/PLATFORM-GUIDE.md` "Workshop" section updated to point at new entry points; underlying scripts still documented

### Code Quality

- [x] Thin-wrapper philosophy held — no duplication of `provision-workshop-roster.sh` logic
- [x] `set -euo pipefail` consistently
- [x] Header docstrings document env vars, flags, and side effects
- [x] Tagged log lines (`[ok]`, `[fail]`, `[skip]`, `[delete]`, `[clean]`) match the existing scripts' conventions
- [x] `shellcheck` clean on all four new scripts (info-level SC2015 cleared up to proper if/then/else in tests)

### Testing

- **Setup tests (17 assertions, 6 scenarios):** happy-path delegation; missing gcloud auth; missing BILLING_ACCOUNT_ID; closed billing account; bad header; empty roster (header-only).
- **Teardown tests (14 assertions, 6 scenarios):** happy path with `--yes`; idempotent re-run; `--yes` doesn't block on closed stdin; bad header rejected; **scope guard** (only roster-derived IDs hit gcloud); failed-delete reflected in summary.
- Tests 5 (teardown) is the most valuable — it forces a regression to "always delete every af-* project" or similar to fail loudly.
- All CI checks pass (lint, test, typecheck, build, python, node, lint-agent-policies, Auto-fix).

### Security

- [x] No command-injection vectors — `gcloud` is argv-parsed; CSV inputs whitespace-stripped before use
- [x] Confirmation prompt is the right safety for destructive ops
- [x] Scope guard via roster-derived IDs limits blast radius
- [x] `roster.csv` (PII) is gitignored from prior PRs; emails in output not logged as credentials

### Recommendation

**GO**

Clean implementation matching the ticket. The thin-wrapper approach is exactly right — the existing provisioning logic stays where it lives, and the new scripts add operational hardening (pre-flight) and safe teardown without touching the hot path. Test 5's whitelist-filter approach to the scope guard is a clever workaround for BSD grep's lack of negative lookahead. Ready to merge.

These are observations, not requested changes. None of them justify a re-spin.

1. **Setup data-row count regex is implicit.** `grep -cE '^[^,]+,'` on line 101 effectively requires "non-empty first column followed by a comma." That's the right semantic but the regex hides it. A one-line comment ("require non-empty handle in column 1") would make it grepable in 6 months.

2. **Teardown's flag parser is tight by design but will need rework if a second flag arrives.** Lines 39–44 accept exactly `--yes` as `$2`. Once we add `--dry-run` or similar (likely soon — the runbook would benefit from a "show me what would happen" mode), refactor to `getopts` or a small while-case loop.

3. **Setup billing check doesn't directly test the empty-stdout path.** Test 4's stub returns `False`; if `gcloud billing accounts list` returned empty (account not visible to this caller), the script would still fall through to `[fail]` correctly, but the test doesn't cover it. Adding a `behavior == "billing-missing"` variant where the stub prints nothing would harden the contract.

4. **Teardown scope guard could harden against a malformed roster.** The guard relies on "IDs come from the CSV" — but if a facilitator accidentally puts `prod` as a handle, the script would compute `af-prod-2026-05` and try to delete it. Adding a regex sanity check on each computed ID (e.g. `^af-[a-z0-9-]+-[0-9]{4}-[0-9]{2}[a-z]?$`) before calling gcloud would close the gap. Edge case, but cheap insurance.

5. **`exec` on line 129 of setup.sh swallows any post-completion summary.** Intentional per PR description; just flagging that any future "lifecycle summary across both pre-flight and provisioning" feature would need to drop the `exec` and capture the inner exit code.

6. **`workshop-teardown.sh` doesn't strip whitespace from `email` while it strips from `handle`/`cohort`.** Email is unused in this script (only handle + cohort form the project ID), so it's fine. Just a small consistency drift from `provision-workshop-roster.sh` which strips all three. Worth aligning if anyone touches both files in the same PR later.

---

tck517 (approved):

PR #23, opened 2026-04-27
fix(scripts): distinguish exists-and-owned from exists-but-not-yours in provisioner skip path

### LINKED ISSUE #22: fix(scripts): distinguish exists-and-owned from exists-but-not-yours in provisioner skip path
## Problem Statement

`scripts/provision-gcp-project.sh`'s "project exists, skip create" branch is too permissive: it checks only whether `gcloud projects describe` returns 0, not whether the caller can actually modify the project. When the project ID exists in **another org** (or was soft-deleted in the caller's own org), `describe` succeeds but every subsequent gcloud call fails with `does not have permission` mid-loop.

Live reproducer 2026-04-27 after re-running the wrapper following a teardown:

```text
[2] bob  ->  af-bob-2026-05
==================================
  Agile Flow GCP Provisioning
==================================
  Project:  af-bob-2026-05
  ...
[skip] Project af-bob-2026-05 already exists
[link] Billing account 0114D1-7670CD-C32F71
ERROR: (gcloud.billing.projects.link) [teddy@vibe.academy] does not have permission to access projects instance [af-bob-2026-05] (or it may not exist): The caller does not have permission.
```

The user thought their teardown had cleaned up cleanly. It hadn't (or the project was never theirs to begin with — GCP project IDs are globally unique). The `[skip]` log line gave a false sense that everything was fine, then gcloud failed three lines later with a generic permission error that doesn't point at the root cause.

This is the **fourth** workshop-blocking provisioning issue (after #14 API-enable, #16 SA-create, #19 org-policy). Different class — not eventual consistency, but a confusing happy-path branch.

**Effort Estimate:** XS
**Priority:** P0 (dry-run is 2026-05-01, 4 days from filing)

---

## A. Environment Context

- Stack: Bash, same as #14/#16/#19
- Files to modify:
  - MODIFY `scripts/provision-gcp-project.sh` — replace bare `describe` check with a probe that verifies modification access
  - MODIFY `scripts/provision-gcp-project.test.sh` — add a scenario for exists-but-not-owned
- Reference:
  - The exact error stderr is captured in the issue Problem Statement
  - GCP project ID uniqueness: <https://cloud.google.com/resource-manager/docs/creating-managing-projects#identifying_projects>

---

## B. Guardrails

- The probe MUST distinguish three states cleanly:
  1. **Doesn't exist** — proceed to create (current behavior)
  2. **Exists and we can modify it** — skip create, continue with billing/APIs/SA (current behavior, just verify perms first)
  3. **Exists but we can't modify it** — fail immediately with a message naming the workaround (change the `cohort` column to generate a fresh ID)
- The probe MUST be a single gcloud call. Do NOT introduce a sequence of probes that takes >5 seconds; the script already has 6 retries × backoff for the eventual-consistency cases.
- Do NOT change the inner script's idempotency contract. Existing-and-owned projects still skip the create step.
- The new failure path MUST exit non-zero with a clear, actionable message. The wrapper's fail-fast behavior should kick in on the first such row, so a partial classroom doesn't happen.
- Do NOT lose the ability to detect deleted-in-own-org projects (`lifecycleState=DELETE_REQUESTED`) — that's a separate skip case that still needs a clear message ("project is in your org but soft-deleted; either restore it or change cohort").

---

## C. Happy Path

The simplest probe: try to read the project's IAM policy. If we have `resourcemanager.projects.getIamPolicy`, we have enough to modify the project (billing-link, IAM bindings, etc.). If we don't, the project isn't ours.

```bash
if gcloud projects describe "$GCP_PROJECT_ID" >/dev/null 2>&1; then
  # Exists. Can we modify it?
  if gcloud projects get-iam-policy "$GCP_PROJECT_ID" >/dev/null 2>&1; then
    state="$(gcloud projects describe "$GCP_PROJECT_ID" --format='value(lifecycleState)')"
    if [[ "$state" == "ACTIVE" ]]; then
      echo "[skip] Project $GCP_PROJECT_ID already exists and is owned by you"
    else
      echo "ERROR: Project $GCP_PROJECT_ID exists in your org but is $state." >&2
      echo "       Either restore it (https://console.cloud.google.com/cloud-resource-manager)" >&2
      echo "       or change the 'cohort' column in roster.csv to generate a fresh ID." >&2
      exit 1
    fi
  else
    echo "ERROR: Project $GCP_PROJECT_ID exists but you do not have permission to modify it." >&2
    echo "       This usually means the ID is taken in another GCP org (project IDs are global)." >&2
    echo "       Change the 'cohort' column in roster.csv to generate a fresh ID." >&2
    exit 1
  fi
elif [[ "$CREATE_PROJECT" == "true" ]]; then
  # ...existing create path...
fi
```

---

## D. Definition of Done

- [ ] `scripts/provision-gcp-project.sh` distinguishes "exists and owned" from "exists but not owned" before the billing-link / IAM / API steps run
- [ ] On "exists but not owned", the script exits 1 with a message naming the cause AND the workaround (change `cohort`)
- [ ] On "exists in your org but not ACTIVE" (e.g. DELETE_REQUESTED), the script exits 1 with a message naming both options (restore or change cohort)
- [ ] On "exists and is ACTIVE in your org", existing skip behavior is preserved
- [ ] On "doesn't exist", existing create path is preserved
- [ ] `scripts/provision-gcp-project.test.sh` gains a scenario stubbing `describe` success + `get-iam-policy` failure → asserts script exits 1 and never calls billing-link
- [ ] All existing 14 helper-test assertions still pass
- [ ] Real-GCP smoke deferred to the 2026-05-01 dry-run; user's `af-bob-2026-05` (currently confused) is the natural smoke target

---

## Why P0

The user just ran into this on real provisioning. Without the fix, every workshop where an `af-{handle}-{cohort}` ID happens to collide globally — or where a participant project from a previous cohort was deleted but the same cohort string is reused — produces this exact mid-loop failure. The error message gcloud emits doesn't point at the cause. Fixing now turns one known-but-confusing failure into a clear, actionable one.

va-worker:
## Summary

Closes #22. **P0 — dry-run is in 4 days.**

Live reproducer today after the user re-ran the wrapper following a teardown:

```text
[skip] Project af-bob-2026-05 already exists
[link] Billing account ...
ERROR: ... does not have permission to access projects instance [af-bob-2026-05]
```

The `[skip]` line is misleading. The project exists somewhere — either in another org (project IDs are globally unique in GCP) or soft-deleted in the user's own org — and the caller has no perms to modify it. Every subsequent gcloud call fails with a confusing generic message.

## What changed

The skip branch now probes ownership via `gcloud projects get-iam-policy`. Four outcomes:

| State | Behavior |
|---|---|
| Doesn't exist | `[create]` (unchanged) |
| Exists + getIamPolicy OK + ACTIVE | `[skip]` (unchanged) |
| Exists + getIamPolicy OK + not ACTIVE | exit 1 with "exists in your org but is in state X" + restore/rename options |
| Exists + getIamPolicy fails | exit 1 with "exists but not yours; project IDs are globally unique" + change-cohort suggestion |

Both failure paths exit BEFORE billing-link, IAM bindings, or any modifying call runs.

## Tests

`scripts/provision-gcp-project.test.sh` gains tests 7 (exists-but-not-yours) and 8 (exists-but-DELETE_REQUESTED). Both stub `gcloud` at PATH and assert:

- Correct exit code (1)
- **`gcloud billing projects link` was never called** — this is the load-bearing assertion. The bug was billing-link firing after a misleading [skip].
- Error stderr contains the cause and the workaround.

All 21 assertions pass (existing 14 retry-helper tests + 7 new ownership/state tests). No regression.

## Implementation notes

- **`set -e` interaction in the test scaffold.** Earlier tests use `set +e/set -e` toggles around their captures, leaving `set -e` ON when the next test starts. Tests 7 and 8 invoke the full provisioner script with stubs that produce expected exit-1 paths; with `set -e` on, the test scaffold would die at that point. Wrapped both new tests in `set +e/set -e`. Worth a future refactor: drop `set -e` from the helper-test entirely and rely on explicit `ec=$?` capture, but that's bigger than this fix.
- **`grep -c` returning 0 for files that don't exist.** Used `[[ -f "$file" ]]` gate plus `grep -c ... || true` rather than the `2>/dev/null || echo 0` pattern, which produces a doubled `0\n0` value under shellcheck-clean redirection. Smaller fix, same outcome.
- **`get-iam-policy` is the right probe.** Cheaper than attempting a no-op IAM binding (no write); has the right semantics (need read access to modify policy); requires the same `resourcemanager.projects.getIamPolicy` permission the script needs anyway. Single gcloud call, no perceptible latency.

## What this does NOT cover

- The user's existing `af-bob-2026-05` is still confused — once this merges, re-running setup with the same roster will produce the new clear error pointing at "change cohort". User then edits roster.csv to use `2026-05a`, re-runs, and provisioning proceeds cleanly.
- This doesn't auto-rename or auto-recover. A facilitator-friendly "rotate cohort suffix" helper could be a follow-up; not in scope here.

## Test plan

- [x] 21/21 helper tests pass
- [x] No regression on existing 14
- [x] CI green (will verify after push)
- [ ] Real-GCP smoke at 2026-05-01 dry-run; `af-bob-2026-05` is the natural target

### DIFF
--- a/scripts/provision-gcp-project.sh
+++ b/scripts/provision-gcp-project.sh
@@ -151,7 +151,30 @@ if [[ "$CREATE_PROJECT" == "true" ]]; then
   fi
 
   if gcloud projects describe "$GCP_PROJECT_ID" >/dev/null 2>&1; then
-    echo "[skip] Project $GCP_PROJECT_ID already exists"
+    # Project ID exists. We need to know whether it's ours and modifiable.
+    # GCP project IDs are global, so an ID can exist in another org and
+    # still respond to `describe`. Probe by attempting to read its IAM
+    # policy — if we have getIamPolicy, we have enough to billing-link
+    # and bind roles. If we don't, the project isn't ours.
+    if gcloud projects get-iam-policy "$GCP_PROJECT_ID" >/dev/null 2>&1; then
+      project_state="$(gcloud projects describe "$GCP_PROJECT_ID" --format='value(lifecycleState)' 2>/dev/null)"
+      if [[ "$project_state" == "ACTIVE" ]]; then
+        echo "[skip] Project $GCP_PROJECT_ID already exists (ACTIVE, owned by you)"
+      else
+        echo "" >&2
+        echo "ERROR: Project $GCP_PROJECT_ID exists in your org but is in state: $project_state" >&2
+        echo "       Either restore it via the Cloud Console (Resource Manager > recently deleted)" >&2
+        echo "       or change the 'cohort' column in roster.csv to generate a fresh project ID." >&2
+        exit 1
+      fi
+    else
+      echo "" >&2
+      echo "ERROR: Project $GCP_PROJECT_ID exists but you do not have permission to modify it." >&2
+      echo "       This usually means the ID is taken in another GCP organization." >&2
+      echo "       (GCP project IDs are globally unique across all of Google Cloud.)" >&2
+      echo "       Change the 'cohort' column in roster.csv to generate a fresh project ID." >&2
+      exit 1
+    fi
   else
     echo "[create] Project $GCP_PROJECT_ID"
     gcloud projects create "$GCP_PROJECT_ID"
--- a/scripts/provision-gcp-project.test.sh
+++ b/scripts/provision-gcp-project.test.sh
@@ -210,6 +210,126 @@ fi
 
 rm -f "$SA_FLAKY" "$COUNTER"
 
+# ── Test 7: project exists but not owned → fail with clear error ─────────
+# Reproduces the live bug observed 2026-04-27 where a roster row hit a
+# project ID that exists outside the caller's reach (different org, or
+# soft-deleted with no perms). describe succeeds, get-iam-policy fails.
+
+echo ""
+echo "Test 7: exists-but-not-yours fails fast with clear message"
+
+T7=$(mktemp -d -t aflowtest-XXXX)
+mkdir -p "$T7/bin"
+
+# Stub gcloud to: describe succeeds, get-iam-policy fails, all else fail.
+cat > "$T7/bin/gcloud" <<EOF
+#!/usr/bin/env bash
+echo "gcloud \$*" >> "$T7/gcloud.log"
+case "\$1 \$2" in
+  "projects describe") exit 0 ;;
+  "projects get-iam-policy")
+    echo "ERROR: PERMISSION_DENIED on getIamPolicy" >&2
+    exit 1
+    ;;
+  *)
+    echo "FATAL: stub should never be called for: \$*" >&2
+    exit 99
+    ;;
+esac
+EOF
+chmod +x "$T7/bin/gcloud"
+
+set +e
+PATH="$T7/bin:$PATH" \
+  GCP_PROJECT_ID="af-collision-2026-05" \
+  BILLING_ACCOUNT_ID="FAKE" \
+  "$SCRIPT" --create-project > "$T7/stdout.log" 2>&1
+ec=$?
+set -e
+
+assert_eq "1" "$ec" "exit 1 on exists-but-not-yours"
+
+# Critical assertion: billing-link must NEVER be called when ownership probe fails.
+# That was the live bug — billing-link fired and produced a confusing error.
+billing_calls=0
+if [[ -f "$T7/gcloud.log" ]]; then
+  billing_calls="$(grep -c 'billing projects link' "$T7/gcloud.log" || true)"
+fi
+assert_eq "0" "$billing_calls" "billing-link was NOT called"
+
+if grep -q "globally unique" "$T7/stdout.log"; then
+  echo -e "  ${GREEN}✓${NC} error message names the global-uniqueness cause"
+  PASS=$((PASS + 1))
+else
+  echo -e "  ${RED}✗${NC} expected error message to mention 'globally unique'; got:"
+  cat "$T7/stdout.log"
+  FAIL=$((FAIL + 1))
+fi
+
+if grep -qi "'cohort' column" "$T7/stdout.log"; then
+  echo -e "  ${GREEN}✓${NC} error message names the workaround"
+  PASS=$((PASS + 1))
+else
+  echo -e "  ${RED}✗${NC} expected error message to suggest changing cohort"
+  FAIL=$((FAIL + 1))
+fi
+
+# ── Test 8: project exists in own org but is DELETE_REQUESTED → fail ─────
+
+echo ""
+echo "Test 8: exists-in-own-org-but-not-ACTIVE fails fast"
+
+T8=$(mktemp -d -t aflowtest-XXXX)
+mkdir -p "$T8/bin"
+
+# Stub: describe succeeds, get-iam-policy succeeds, lifecycleState query
+# returns DELETE_REQUESTED.
+cat > "$T8/bin/gcloud" <<EOF
+#!/usr/bin/env bash
+echo "gcloud \$*" >> "$T8/gcloud.log"
+case "\$1 \$2" in
+  "projects describe")
+    # The script calls describe twice: once to check existence (no flags),
+    # once to read lifecycleState. Distinguish by --format presence.
+    if echo "\$*" | grep -q -- "--format"; then
+      echo "DELETE_REQUESTED"
+    fi
+    exit 0
+    ;;
+  "projects get-iam-policy") exit 0 ;;
+  *)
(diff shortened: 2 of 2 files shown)


va-reviewer:
## PR Review — #23

### Requirements

- [x] Distinguishes exists-and-owned from exists-but-not-owned BEFORE billing-link
- [x] Exists-but-not-owned exits 1 with cause (global-uniqueness) + workaround (change cohort)
- [x] Exists-but-not-ACTIVE exits 1 with both options (restore or rename)
- [x] Exists-and-ACTIVE preserves skip behavior
- [x] Doesn't-exist preserves create path
- [x] Test scenario covers exists-but-not-owned (Test 7); bonus Test 8 covers DELETE_REQUESTED
- [x] No regression on existing 14 retry-helper assertions (CI green)

### Code Quality

- [x] Four-state decision tree enumerated cleanly: doesn't exist / exists+owned+ACTIVE / exists+owned+other / exists+not-owned
- [x] Error messages name **both** cause and workaround — exactly the actionability the live bug was missing
- [x] Comment explains *why* `get-iam-policy` is the probe (read access implies write capability for our needs)
- [x] No unnecessary complexity — minimal new logic, just a bracketing if/else around the existing skip path
- [x] Correctly handles the `set -e` shell-state-leak (recurring lesson from #16) by wrapping test invocations in `set +e/set -e`

### Testing

- **Test 7 (exists-but-not-yours):** stubs gcloud's `describe` success + `get-iam-policy` failure; asserts exit 1, **billing-link never called**, error names cause + workaround. The stub's `*)` arm is defensive — emits FATAL and exits 99 if the script regresses to calling any other gcloud subcommand on the failure path. Strong coverage.
- **Test 8 (DELETE_REQUESTED):** clever stub differentiates the two `describe` calls (existence check vs `--format=lifecycleState`) by argument presence. Tightly coupled to the script's call pattern, but the coupling is explicitly commented.
- 21/21 assertions pass (existing 14 + 7 new across two scenarios).
- All CI checks pass.

### Security

- [x] No command-injection vectors — `gcloud` argv-parsed; `$GCP_PROJECT_ID` flows through unmodified
- [x] Error messages don't leak credentials or participant PII
- [x] Stub's `*)` FATAL arm prevents test from accidentally triggering real gcloud calls

### Recommendation

**GO**

P0 fix for a live failure observed today on `af-bob-2026-05`. The probe-via-`get-iam-policy` approach is correct, the failure-path error messages are actionable (name both cause and workaround), and tests assert on the load-bearing invariant: `billing-link` must never run when the ownership probe fails. That's exactly what the bug was — billing-link fired after a misleading `[skip]` and produced a generic permission error. With this fix, the same scenario surfaces a clear "exists but not yours; change cohort" message before any modifying call runs. Ready to merge.

These are observations from a careful read; none justify a re-spin.

1. **`project_state` could be empty if the second `describe` fails.** If `get-iam-policy` succeeds but the immediately-following `describe --format='value(lifecycleState)'` returns empty (race? quota hiccup?), the script falls through to the "not ACTIVE" branch with state shown as `''`. Confusing message but not a safety issue. A tiny fix: `if [[ -z "$project_state" ]]; then echo "could not determine project state"; exit 1; fi` before the `[[ ... == "ACTIVE" ]]` check.

2. **Two `describe` calls could be one.** Lines 153 and 158 both query the project. Combining into a single `gcloud projects describe ... --format='value(lifecycleState)'` (treating empty stdout / non-zero exit as "doesn't exist", non-empty as "exists with this state") would cut one round-trip per row. ~30% latency improvement on the existence-check phase. Not worth respinning, but a clean follow-up.

3. **`get-iam-policy` permission ≠ billing-link permission, in theory.** A read-only viewer would pass the probe but fail billing-link. In practice, workshop facilitators are project creators with `roles/owner`, so this isn't a real concern. The comment on line 154 says "if we have getIamPolicy, we have enough to billing-link" — strictly that's "probably enough." Worth softening the comment if anyone touches this block again.

4. **Test 8's `--format` discriminator is tightly coupled.** Lines 277-281 distinguish the two `describe` calls by checking for `--format` in the args. If the script ever adds `--format` to the existence check (e.g., for caching), the stub's logic silently breaks. Already commented, so future-you will see it.

5. **No direct test for the unchanged happy path** (project exists, owned, ACTIVE → skip and continue). The existing tests don't exercise the full provisioner flow either, so this is a pre-existing gap rather than a regression risk specific to this PR. A Test 9 stubbing the full happy-skip path would round out coverage. Low priority — the regression risk is low because the fix only modifies the "exists" branch and ACTIVE is the simplest case.

---

PR #24, opened 2026-04-27
fix(scripts): override iam.allowedPolicyMemberDomains in provisioner; correct doc command

### LINKED ISSUE #19: fix(scripts): override iam.allowedPolicyMemberDomains in provisioner; correct doc command

va-worker:
## Summary

Closes #19. **P0 — dry-run is in 4 days.**

Two fixes that ship together:

### 1. provision-gcp-project.sh — new Step 1.5

After project creation and billing-link, before APIs/SA/bindings, the script now detects whether the parent org enforces `iam.allowedPolicyMemberDomains` and applies a per-project override (`allValues: ALLOW`) if so. Three branches:

| Branch | Log |
|---|---|
| Override already in place (`allValues=ALLOW`) | `[skip] override already in place` |
| Constraint enforced, override not yet applied | `[override] applying domain-restricted-sharing override` |
| Constraint absent | `[skip] not enforced; no override needed` |

Idempotent — re-runs through the first branch.

### 2. PATTERN-LIBRARY.md pattern #30 — correct gcloud command

Previous version used `disable-enforce`, which **only works on boolean constraints**. `iam.allowedPolicyMemberDomains` is a list constraint; `disable-enforce` against it returns `INVALID_ARGUMENT: Policy and Constraint must be of the same type`. The correct invocation is `set-policy` with the explicit `listPolicy: { allValues: ALLOW }` body. Pattern #30 now reflects that, with a callout warning future readers about the trap.

The pattern's "Where this fits" section is updated to note the override is now automatic per-project; the manual snippet remains for recovery scenarios.

## Tests

`scripts/provision-gcp-project.test.sh` gains tests 9–11 covering all three branches:

- Test 9: override already applied → `[skip]` message, `set-policy` NOT called
- Test 10: constraint enforced → `set-policy` called exactly once, body contains `"allValues":"ALLOW"`
- Test 11: constraint absent → `[skip]` message, `set-policy` NOT called

All 28 assertions pass (existing 21 + 7 new). No regression.

## Companion PR

`agile-flow-meta`#TBD — runbook §5 also has the wrong `disable-enforce` command; will be filed alongside this. Same fix shape: replace with `set-policy`, mark Step 1.5 as automatic.

## Test plan

- [x] `bash -n` and `shellcheck` clean on the script and the test
- [x] 28/28 helper tests pass
- [x] No regression on existing 21
- [x] markdownlint clean on PATTERN-LIBRARY.md
- [ ] Real-GCP smoke at the 2026-05-01 dry-run

## Implementation notes

- **Detection uses two gcloud calls.** First, `org-policies describe ... --format='value(listPolicy.allValues)'` — if it prints `ALLOW`, the override is already in place. Otherwise, `org-policies list` is grepped for the constraint name; if present, set-policy runs. If neither, skip. This avoids re-applying when the policy is already in the right state, which would otherwise be a wasted `set-policy` call per re-run.
- **set-policy body via `/dev/stdin`.** The `echo '{...}' | gcloud ... set-policy /dev/stdin` form works in both bash and fish; no heredoc indentation traps. Same shape the user discovered manually in fish.
- **Step 1.5 placement.** After project creation + billing-link (Step 1) and before API enable (Step 2). The deployer-SA bindings later target a service account on the project's own domain so they don't actually hit the constraint, but the *participant* binding in `provision-workshop-roster.sh` does — so applying the override early is the correct seam.

### DIFF
--- a/docs/PATTERN-LIBRARY.md
+++ b/docs/PATTERN-LIBRARY.md
@@ -993,24 +993,35 @@ first IAM binding for the very first participant fails.
 
 **Pattern: disable the constraint per-project, scoped to workshop projects only.**
 
-Run this *once per participant project*, after `provision-gcp-project.sh`
-creates the project and before `gcloud projects add-iam-policy-binding`
-attempts the binding (the workshop wrapper calls the binding directly,
-so the override must run in between):
+As of `provision-gcp-project.sh` Step 1.5, the override is applied
+automatically per project right after creation. You do not normally need
+to run it by hand. The pattern below is for reference and for
+out-of-band cases (e.g. recovering an existing project).
+
+> **`disable-enforce` does not work here.** This is a *list constraint*
+> (with `allowedValues` / `allValues`), not a boolean. `disable-enforce`
+> only works on boolean constraints — using it returns
+> `INVALID_ARGUMENT: Policy and Constraint must be of the same type`.
+> Use `set-policy` with the YAML/JSON `listPolicy: { allValues: ALLOW }`
+> form below.
 
 ```bash
-# Check whether the org has the constraint enforced. Idempotent — no-op
-# if not enforced. Requires roles/orgpolicy.policyAdmin on the project,
+# Apply per project. Idempotent — re-running is a no-op.
+# Requires roles/orgpolicy.policyAdmin on the project,
 # which the project creator has by default.
 
 if gcloud resource-manager org-policies describe \
   iam.allowedPolicyMemberDomains \
   --project="$GCP_PROJECT_ID" \
-  --format='value(booleanPolicy.enforced)' 2>/dev/null | grep -q true; then
-  echo "[override] disabling domain-restricted-sharing for $GCP_PROJECT_ID"
-  gcloud resource-manager org-policies disable-enforce \
-    iam.allowedPolicyMemberDomains \
-    --project="$GCP_PROJECT_ID"
+  --format='value(listPolicy.allValues)' 2>/dev/null | grep -q '^ALLOW$'; then
+  echo "[skip] override already in place for $GCP_PROJECT_ID"
+elif gcloud resource-manager org-policies list \
+  --project="$GCP_PROJECT_ID" \
+  --format='value(constraint)' 2>/dev/null | grep -q 'allowedPolicyMemberDomains'; then
+  echo "[override] applying domain-restricted-sharing override for $GCP_PROJECT_ID"
+  echo '{"constraint":"constraints/iam.allowedPolicyMemberDomains","listPolicy":{"allValues":"ALLOW"}}' \
+    | gcloud resource-manager org-policies set-policy /dev/stdin \
+        --project="$GCP_PROJECT_ID"
 fi
 ```
 
@@ -1036,8 +1047,10 @@ that failed will succeed within seconds. There is no propagation lag
 on the policy override itself in our experience, but the SA-binding
 retry helper covers the rare case where there is.
 
-**Where this fits in the workshop flow:** the runbook's T-3 days
-provisioning step covers the override. Once
-`provision-workshop-roster.sh` ships the override inline (planned
-follow-up; not yet shipped), facilitators won't need to think about
-this.
+**Where this fits in the workshop flow:** as of #19,
+`provision-gcp-project.sh` runs the override automatically per project
+(Step 1.5, between project creation and API enable). Facilitators do
+not need to apply it manually for new workshop projects. The runbook's
+T-3 days section retains a recovery snippet for projects provisioned
+before #19, or for cases where the override needs to be re-applied
+out-of-band.
--- a/scripts/provision-gcp-project.sh
+++ b/scripts/provision-gcp-project.sh
@@ -185,6 +185,35 @@ if [[ "$CREATE_PROJECT" == "true" ]]; then
     --billing-account="$BILLING_ACCOUNT_ID"
 fi
 
+# ── Step 1.5: Override Domain Restricted Sharing (workshop projects) ─────
+#
+# If the parent org enforces `iam.allowedPolicyMemberDomains` (a list
+# constraint, on by default for Cloud Identity Free orgs), every
+# subsequent IAM binding to an external-domain identity (Gmail, other
+# Workspace) fails with FAILED_PRECONDITION. Override per project:
+# allValues=ALLOW so any Google identity can be a binding member.
+# Production projects in the same org keep their constraint.
+#
+# This is a list constraint, NOT a boolean — `disable-enforce` does not
+# work; `set-policy` with the explicit listPolicy form does. See
+# docs/PATTERN-LIBRARY.md pattern #30.
+
+if gcloud resource-manager org-policies describe \
+  iam.allowedPolicyMemberDomains \
+  --project="$GCP_PROJECT_ID" \
+  --format='value(listPolicy.allValues)' 2>/dev/null | grep -q '^ALLOW$'; then
+  echo "[skip] domain-restricted-sharing override already in place for $GCP_PROJECT_ID"
+elif gcloud resource-manager org-policies list \
+  --project="$GCP_PROJECT_ID" \
+  --format='value(constraint)' 2>/dev/null | grep -q 'allowedPolicyMemberDomains'; then
+  echo "[override] applying domain-restricted-sharing override for $GCP_PROJECT_ID"
+  echo '{"constraint":"constraints/iam.allowedPolicyMemberDomains","listPolicy":{"allValues":"ALLOW"}}' \
+    | gcloud resource-manager org-policies set-policy /dev/stdin \
+        --project="$GCP_PROJECT_ID"
+else
+  echo "[skip] domain-restricted-sharing not enforced; no override needed"
+fi
+
 # ── Step 2: Enable APIs ──────────────────────────────────────────────────
 
 echo "[enable] Required APIs (may take 30-60 seconds on first run)"
--- a/scripts/provision-gcp-project.test.sh
+++ b/scripts/provision-gcp-project.test.sh
@@ -330,6 +330,153 @@ else
   FAIL=$((FAIL + 1))
 fi
 
+# ── Test 9-11: Step 1.5 domain-restricted-sharing override ──────────────
+#
+# The script's Step 1.5 has three branches:
+#   - override already in place (allValues=ALLOW)  → [skip]
+#   - constraint enforced, override not yet applied → set-policy is called
+#   - constraint not enforced at all                → [skip]
+#
+# Each branch is exercised below. Stubs inject controlled responses to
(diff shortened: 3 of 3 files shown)


va-reviewer:
## PR Review — #24

### Requirements

- [x] Step 1.5 detects + applies + skips correctly across three states (already-applied / enforced+not-applied / not-enforced)
- [x] Override is project-scoped only — `--project=$GCP_PROJECT_ID` on every gcloud call
- [x] Detection is two gcloud calls worst-case; no perceptible latency added (well under the ticket's 5-second budget)
- [x] Idempotency preserved — already-applied state takes `[skip]` branch
- [x] Distinguishes "not enforced" from "enforced + applied" cleanly (three branches, not two)
- [x] Runs **after** project creation, **before** API enable + SA + bindings — correct seam
- [x] Pattern #30 corrected: `disable-enforce` → `set-policy` with explicit `listPolicy` form
- [x] Warning callout added in pattern #30 about the list-vs-boolean trap that bit the user live
- [x] Pattern #30's "Where this fits" updated to note the override is now automatic
- [x] Test scenarios cover all three Step 1.5 branches
- [x] No regression on existing 21 assertions

### Code Quality

- [x] Step 1.5 is 22 lines including the docstring comment — minimal new logic
- [x] Inline comment explains the list-vs-boolean trap so a future reader doesn't try `disable-enforce`
- [x] `grep -q '^ALLOW$'` is anchored — catches exactly the ALLOW token, not partial matches
- [x] `echo '{...}' | gcloud ... set-policy /dev/stdin` avoids heredoc indentation traps (fish, zsh)
- [x] No unnecessary complexity — no retry wrap, no error-classification scaffold; Step 1.5 fails-fast like the rest of the script
- [x] Clear log line per branch so a facilitator can see what happened during real provisioning

### Testing

- **Tests 9–11** cover all three Step 1.5 branches via a `run_step1_5_test` helper that stubs gcloud's three relevant subcommands.
- **Clever short-circuit:** the stub returns non-zero on `services enable` so the script halts at Step 2. Step 1.5 has fully executed by then, and the assertions read stdout + gcloud-call log to verify behavior. Documented in a comment block at the top of the test section.
- **Strong assertions:** Test 10 doesn't just check that `set-policy` was called — it asserts the JSON body contains `"allValues":"ALLOW"`. That guards against a regression where the script calls set-policy with the wrong policy shape.
- All 28 assertions pass (existing 21 + 7 new across 3 scenarios).
- All CI checks pass.

### Security

- [x] The JSON policy body is a static literal, not user-controlled — no injection vector
- [x] Override broadens IAM membership **for the workshop project only** — production projects in the same org keep the constraint. Pattern #30 documents this tradeoff explicitly.
- [x] Test stubs isolated via `mktemp -d`

### Recommendation

**GO**

P0 closure of the manual workaround. The three-branch decision is correctly enumerated, the gcloud command shape is the one the user proved works manually, and the test asserts on the JSON body shape (not just call count) — which is the load-bearing assertion for "the script is doing the right thing, not just calling the right command name." Ready to merge.

Suggested merge order: this PR first, then `agile-flow-meta`#88, so the runbook's "as of #19" claim is true on main.

1. **`run_step1_5_test`'s `label` parameter is unused.** Line 55 of the test diff declares `local label="$1"` but never references it. Either drop it or use it in error messages — the assertions could prefix with `[$label]` for nicer multi-test output. shellcheck didn't catch it because the `local x=...` form looks like an assignment.

2. **Consider wrapping Step 1.5 calls in `retry_eventual_consistency`** for symmetry with Steps 2/4/5. The org-policy API has its own propagation behavior; if eventual-consistency ever hits a fresh project's policy API, the current Step 1.5 silently treats the failure as "not enforced" and skips. No evidence of this hitting workshops yet, so wait-and-see is fine.

3. **Test 12 for set-policy permission failure** would lock in the desired behavior (`set -e` halts; clear gcloud error surfaces). Today the path is correct but untested — a future refactor to swallow the error would silently regress. One-shot test stubbing `set-policy` to exit 1 with a permission message.

4. **The pattern-library "alternatives considered and rejected" section is unchanged.** All three alternatives are still valid. Nothing to flag — just noting that's intentional given the pattern's evergreen content.

5. **Cross-link from Step 1.5's inline comment to pattern #30** is already there (line 184 of the script diff). Good craft. Future reader gets the full story without leaving the script.

---

PR #26, opened 2026-04-27
feat(scripts): automate Workload Identity Federation per project

va-worker:
## Summary

Closes #5. Automates the last manual gcloud step in the provisioning flow.

`provision-gcp-project.sh` gains Step 5.5: when `GITHUB_USERNAME` is set, the script creates the WIF pool, the OIDC provider, and the IAM binding scoped to `${GITHUB_USERNAME}/${WIF_REPO:-agile-flow-gcp}`. The workshop wrapper exports `GITHUB_USERNAME` automatically from each `roster.csv` row, so a facilitator running the canonical workshop flow never touches WIF gcloud commands by hand.

The four GitHub Actions secrets a participant pastes into their fork now fall out of provisioner output directly — no copying from docs, no project-number arithmetic.

## Scope changes vs. original ticket

(Already documented on the issue body — I'll mention them here for review context.)

| Original ticket | This PR |
|---|---|
| `--with-wif owner/repo` flag | `GITHUB_USERNAME` env var (matches existing env-var pattern) |
| Implicit `repo` column in roster | No CSV schema change; repo hardcoded to `agile-flow-gcp` with `WIF_REPO` env override |
| Real-GCP end-to-end verification | Mocked-shell tests + dry-run deferral (consistent with #4/#14/#16/#19/#22) |

User explicitly approved the hardcoded-repo decision: workshop participants fork without renaming, so `<github_user>/agile-flow-gcp` is the de-facto convention. `WIF_REPO=` stays as the escape hatch.

### `scripts/provision-gcp-project.sh`

New Step 5.5 (~70 lines) between the deployer-SA IAM bindings and the optional `--with-sa-key` path. Three sub-steps (all idempotent):

- 5.5a: Create WIF pool `github` if absent
- 5.5b: Create OIDC provider `github` if absent
- 5.5c: Bind `roles/iam.workloadIdentityUser` on the deployer SA to `attribute.repository/<github_user>/<wif_repo>`

The "Next steps" output at the end of the script now prints the concrete WIF provider resource path (`projects/<number>/locations/...`) instead of a placeholder, so the facilitator copy-pastes directly into participant fork secrets.

### `scripts/provision-workshop-roster.sh`

- Reads `github_user` from each CSV row, strips whitespace, exports `GITHUB_USERNAME` before calling the inner script
- After each row, queries the project number and writes a concrete `wif_provider` value into `roster-output.csv` (was empty placeholder before)
[...]

PR #27, opened 2026-04-28
fix(scripts): add required --attribute-condition to WIF OIDC provider

va-worker:
## Summary

Quick fix — no linked ticket. Live failure during workshop smoke just now:

```text
ERROR: (gcloud.iam.workload-identity-pools.providers.create-oidc) INVALID_ARGUMENT:
The attribute condition must reference one of the provider's claims.
```

Google now requires `--attribute-condition` on OIDC providers. PR #26 removed it on the (logically correct) assumption that the IAM binding gates access — but Google rejects provider creation entirely without a condition.

## Fix

Use a trivially-true condition:

```bash
--attribute-condition="assertion.repository != ''"
```

Always passes for any GitHub OIDC token. Satisfies Google's "must reference a claim" rule. **Doesn't restrict access** by org or repo — trust scoping stays at the IAM binding layer where `attribute.repository=<user>/<repo>` pins each binding to one specific repo.

## What changed

- **`scripts/provision-gcp-project.sh`** Step 5.5b: add the condition; updated inline comment to explain the choice
- **`docs/PLATFORM-GUIDE.md`** manual-fallback snippet: same condition so copy-paste users hit the same fix
- **`scripts/provision-gcp-project.test.sh`** test 13: assert the `create-oidc` invocation includes `--attribute-condition`. Future-proofs against someone removing it again.

## Why this slipped past mocked tests

The mocked-shell tests verified the script *issues* the create-oidc call. They didn't assert what *flags* it passed. Real gcloud rejected the flagless call; a `gcloud projects describe` stub couldn't have caught that. The new assertion in test 13 closes the gap — any future regression to "no condition" fails the test before it ships.

This is the second test-mismatch lesson from the workshop work (the first was the bash-quote-stripping bug in #26). Mocked tests verify call shape; they can't verify call *acceptance* by the real API. Real-GCP smoke at the dry-run remains the load-bearing acceptance signal.

## Test plan

- [x] 38/38 helper tests pass (existing 37 + 1 new condition assertion)
- [x] No regression on wrapper
- [x] markdownlint clean
- [ ] Real-GCP re-run on `af-reviewer-2026-05a` / `af-worker-2026-05a` — that's what surfaced the bug

### DIFF
--- a/docs/PLATFORM-GUIDE.md
+++ b/docs/PLATFORM-GUIDE.md
@@ -200,6 +200,7 @@ gcloud iam workload-identity-pools providers create-oidc github \
   --location=global \
   --issuer-uri="https://token.actions.githubusercontent.com" \
   --attribute-mapping="google.subject=assertion.sub,attribute.repository=assertion.repository,attribute.actor=assertion.actor" \
+  --attribute-condition="assertion.repository != ''" \
   --project=YOUR_PROJECT_ID
 
 # Get the project number (different from project ID)
--- a/scripts/provision-gcp-project.sh
+++ b/scripts/provision-gcp-project.sh
@@ -301,11 +301,14 @@ done
 # <github_user>/agile-flow-gcp. Override WIF_REPO if you ever need a
 # different repo name (dry-run smoke, non-workshop callers).
 #
-# We deliberately do NOT add --attribute-condition restricting to a
-# specific GitHub org — workshop participants fork under their personal
-# accounts, not the canonical org. Access is scoped at the IAM binding
-# layer instead, where attribute.repository=<user>/agile-flow-gcp
-# pins the trust to one specific repo.
+# Google requires --attribute-condition on OIDC providers (it must
+# reference at least one provider claim). We use a trivially-true
+# condition (`assertion.repository != ''`) so the provider doesn't gate
+# access by org or repo — workshop participants fork under their
+# personal GitHub accounts, not the canonical org, and we cannot
+# enumerate every domain ahead of time. Trust scoping happens at the
+# IAM binding layer instead, where attribute.repository=<user>/<repo>
+# pins each binding to one specific repo.
 #
 # All three sub-steps (pool, provider, binding) are idempotent.
 
@@ -342,6 +345,7 @@ if [[ -n "${GITHUB_USERNAME:-}" ]]; then
       --location=global \
       --issuer-uri="https://token.actions.githubusercontent.com" \
       --attribute-mapping="google.subject=assertion.sub,attribute.repository=assertion.repository,attribute.actor=assertion.actor" \
+      --attribute-condition="assertion.repository != ''" \
       --project="$GCP_PROJECT_ID"
   fi
 
--- a/scripts/provision-gcp-project.test.sh
+++ b/scripts/provision-gcp-project.test.sh
@@ -620,6 +620,17 @@ else
   FAIL=$((FAIL + 1))
 fi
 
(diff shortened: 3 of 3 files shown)


va-reviewer:
## PR Review — #27

### Requirements

- [x] OIDC provider creation now satisfies Google's "must reference a claim" rule via `--attribute-condition="assertion.repository != ''"`
- [x] Condition is trivially-true — does not restrict access by org or repo
- [x] Manual fallback in PLATFORM-GUIDE.md updated identically so copy-paste users hit the same fix
- [x] Inline comment rewritten to explain *why* the condition exists and why it's permissive
- [x] Test 13 extended to assert the `create-oidc` invocation includes `--attribute-condition`
- [x] No regression — 38/38 helper tests pass; CI green

### Code Quality

- [x] Three-line code change + one updated comment + one test assertion. Surgical scope appropriate for a quick fix to a live failure.
- [x] Trust scoping argument unchanged from PR #26 — the IAM binding's `attribute.repository=<user>/<repo>` is still the access gate
- [x] No new abstractions, no hardcoded values that should be configurable
- [x] `shellcheck` clean

### Testing

- [x] **New assertion in Test 13** asserts the `gcloud workload-identity-pools providers create-oidc` invocation includes `--attribute-condition`. Future-proofs against someone removing it again.
- [x] Test 14 (idempotent path) is unaffected — `create-oidc` doesn't run when artifacts already exist, and the new assertion is correctly scoped to test 13.
- [x] All CI checks pass.

### Security

- [x] Static-literal condition, no user-input injection vector
- [x] Net security posture unchanged from PR #26's design intent — the IAM binding gates per-repo access; the condition is a one-claim check that nominally broadens nothing in practice
- [x] No secrets, no `--no-verify`, no bypassed hooks

### Recommendation

**GO**

Real bug, minimal fix, the new assertion future-proofs against regression. The choice of `assertion.repository != ''` over alternatives (`repository_owner == 'foo'`, prefix matchers) is correctly aligned with the workshop's "participants fork to their own accounts" model.

The lesson noted in the PR description — *mocked tests verify call shape; they can't verify call acceptance by the real API* — is worth keeping. The new test assertion at least closes the "shape" gap (the flag is now mandatory). Real-GCP smoke at the dry-run remains the only acceptance signal for the actual gcloud-server contract.

These are observations from a careful read; none justify a re-spin.

1. **Test asserts presence, not value.** `grep -q "providers create-oidc.*attribute-condition"` matches any `--attribute-condition` value, including a hypothetical `="false"` that would reject all tokens. Tightening to `grep -q "attribute-condition='assertion.repository != ''"` would lock in the specific value. Trade-off: ties the test to today's exact wording, which the team might want to evolve. Today's looser check is reasonable.

2. **Stylistic alternatives** — `assertion.aud != ''` or `assertion.iss != ''` would also satisfy Google's rule. They read slightly more clearly as "this is a satisfaction condition, not a meaningful filter" since `repository` is the field that *does* gate access at the binding layer. Pure style; the current form is fine.

3. **Defense-in-depth angle.** `assertion.repository != ''` technically asserts the token came from a repo-scoped GitHub Actions context (not, say, a personal-token impersonation that happens to have an OIDC token). Tiny defensive value — not worth highlighting in the comment, but worth knowing if you ever consider further hardening.

4. **The "real-GCP smoke is the load-bearing signal" lesson** is now confirmed twice this session (the bash-quote-stripping bug in #26, and now this). Worth a one-line note in the testing patterns area of CLAUDE.md or PATTERN-LIBRARY.md so future PRs to gcloud-calling scripts know to expect this gap. Not in scope here.

---

PR #29, opened 2026-04-28
fix(ci): remove secrets.* from if: expressions in deploy/preview/rollback

### LINKED ISSUE #28: fix(ci): remove secrets.* from if: expressions in deploy/preview workflows
## Problem Statement

Three CI/CD workflows in this repo reference `secrets.*` inside `if:` expressions on individual steps. **GitHub Actions hard-rejects this pattern at workflow load time** — the run records as `failure` with `0s` elapsed and "log not found", and no jobs execute. Every fork that pushes to `main` hits this on the first deploy.

Live reproducer 2026-04-28: `tck517/agile-flow-gcp` (a fork of this repo) pushed to `main`, `deploy.yml` recorded a `0s` startup_failure, no log, no Cloud Run service created. The exact YAML lines that trigger the rejection:

```yaml
# deploy.yml:51, 58
if: steps.check_secrets.outputs.skip != 'true' && secrets.GCP_WORKLOAD_IDENTITY_PROVIDER != ''
if: steps.check_secrets.outputs.skip != 'true' && secrets.GCP_WORKLOAD_IDENTITY_PROVIDER == '' && secrets.GCP_SA_KEY != ''
```

GitHub blocks `secrets.*` in `if:` to prevent secret values from leaking through expression evaluation. The rule has been in place since at least 2022; this is a real GitHub validator behavior, not a transient bug. Authoritative reference: <https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability> (the `secrets` context is unavailable in `if:` conditionals on steps and jobs except in `secrets.GITHUB_TOKEN` cases).

The fix is to **promote secret-presence checks into a previous step's outputs**, then gate downstream steps on `steps.X.outputs.Y` (which `if:` is allowed to read).

This is a **workshop-blocking** bug. Every participant will hit it on the first deploy of their fork. The dry-run for the May 6 workshop is on May 1 (3 days from filing).

**Effort Estimate:** S (template-wide search-and-replace; one new step output, three workflows)
**Priority:** P0

---

## A. Environment Context

- Stack: GitHub Actions YAML
- Files to modify:
  - MODIFY `.github/workflows/deploy.yml` — 2 secrets-in-if references; existing `check_secrets` step needs an `auth_method` output
  - MODIFY `.github/workflows/preview-deploy.yml` — 3 secrets-in-if references; existing `check_secrets` step needs additional outputs (Neon presence, auth method)
  - MODIFY `.github/workflows/preview-cleanup.yml` — 5 secrets-in-if references; no `check_secrets` step exists; one needs to be added
- Reference:
  - GitHub Actions context-availability rules (link above)
  - The existing `check_secrets` step in `deploy.yml` is the model — it already promotes `skip` to an output successfully. We're extending the same pattern.

---

## B. Guardrails

- Do NOT change the *behavior* of the workflows. The same secrets-presence logic must apply; only the *expression site* moves from `if:` to `run:`.
- Do NOT introduce new secret names. The set of secrets remains: `GCP_PROJECT_ID`, `GCP_WORKLOAD_IDENTITY_PROVIDER`, `GCP_SERVICE_ACCOUNT`, `GCP_SA_KEY`, `NEON_API_KEY`, `NEON_PROJECT_ID`, `PRODUCTION_DATABASE_URL`.
- Each promoted output MUST encode a single decision (`skip` / `auth_method` / `neon_configured`). Avoid stuffing multiple booleans into one output — debugging gets harder.
- The `check_secrets` step pattern MUST be added to `preview-cleanup.yml` (it has none today). Don't try to inline the bash check — `if:` would still fail on the inlined step.
- Do NOT add tests for these changes. Workflow YAML is verified by GitHub Actions on push; we don't have a local YAML linter that catches the secrets-in-if rule. Real-fork smoke (the user's `tck517/agile-flow-gcp` re-run after the fix lands) is the acceptance signal.

---

## C. Happy Path

1. PR merges. Workflow YAML is now valid for forks.
2. User nukes `tck517/agile-flow-gcp` and re-forks `vibeacademy/agile-flow-gcp` from a clean state.
3. User adds the four GitHub Actions secrets (or re-uses prior values).
4. User pushes a trivial change to `main`.
5. `deploy.yml` runs successfully: `check_secrets` step computes `auth_method=wif`, the WIF auth step runs, build + push + deploy proceed.
6. Cloud Run service is created; `gcloud run services list` returns one row.
7. `/api/health` returns 200.

---

## D. Definition of Done

- [ ] `.github/workflows/deploy.yml`: `check_secrets` step emits an `auth_method` output (`wif` | `sa-key` | `none`); both auth steps gate on `steps.check_secrets.outputs.auth_method == '<method>'`. No `secrets.*` references remain in any `if:` expression.
- [ ] `.github/workflows/preview-deploy.yml`: same shape. The Neon-conditional steps gate on a new `neon_configured` output, not on `secrets.NEON_API_KEY != ''`.
- [ ] `.github/workflows/preview-cleanup.yml`: a new `check_secrets` step is added at the top of the job. All downstream steps gate on its outputs.
- [ ] `gh workflow view <name> --yaml` for each modified workflow shows zero `secrets.X` references in `if:` lines (verifiable via `grep "if:.*secrets\." .github/workflows/*.yml` returning empty).
- [ ] No behavioral change — the same skip-when-not-configured semantics hold.
- [ ] Real-fork smoke deferred to user's re-run on `tck517/agile-flow-gcp` after merge.

---

## Why P0

This is the second-to-last gate between "scripts work" and "participant deploys their first push to Cloud Run." Every fork will hit this on first deploy without the fix. Three days to the May 1 dry-run. If the dry-run can't reach a working Cloud Run service, every subsequent fix won't help.

---

## Notes on the test scaffolding

We learned this session that mocked-shell tests verify call shape but not call acceptance by the real API. Workflow YAML has the same gap — the validator only runs on GitHub's side. Worth adding a follow-up ticket for a local lint rule (`actionlint`?) that catches `secrets.*` in `if:` before push. Not in scope here.

va-worker:
## Summary

Closes #28. **P0 — workshop dry-run is in 3 days.**

GitHub Actions hard-rejects `secrets.*` in `if:` expressions. Workflows that violate the rule fail at workflow-load time with `0s` elapsed, no logs, no jobs run. **Every fork hits this on first deploy.**

Live reproducer this afternoon: `tck517/agile-flow-gcp` (fresh fork) pushed to `main`, `deploy.yml` recorded a `0s` startup_failure. No Cloud Run service was ever created. The same pattern would block every workshop participant on day-1 step 4.

## What changed

Promote secret-presence checks into a `check_secrets` step's outputs in each affected workflow. Downstream auth and conditional steps gate on `steps.check_secrets.outputs.X` (allowed in `if:`) instead of `secrets.X` (rejected).

Per-workflow:

| Workflow | Output(s) added | Notes |
|---|---|---|
| `deploy.yml` | `auth_method` (wif / sa-key / none) | Existing `check_secrets` step extended |
| `preview-deploy.yml` | `auth_method`, `neon_configured` | Existing `check_secrets` step extended |
| `preview-cleanup.yml` | `auth_method`, `gcp_configured`, `neon_configured` | **New `check_secrets` step** — none existed before |
| `rollback-production.yml` | `auth_method` | Existing step extended; hard-fails if no auth method available |

No behavioral change — same skip semantics, same auth precedence (WIF over SA key when both set).

## Verification

```bash
grep -nE "if:.*[^_]secrets\.[A-Z]" .github/workflows/*.yml
```

Returns zero matches (the negative lookbehind `[^_]` excludes the `check_secrets` step ID, which contains the substring "secrets" but is not the `secrets` context).

## Why mocked tests didn't catch this

Workflow YAML is validated by GitHub Actions on push. We have no local linter for the secrets-in-if rule. **`actionlint` does catch it** — worth a follow-up ticket to add a pre-push hook or CI step that runs actionlint on `.github/workflows/*.yml`.

This is the third "mocked tests verify shape, not acceptance" lesson this session:
- #26 had a bash-quote-stripping bug
- #27 was missing `--attribute-condition` (Google requires it)
- This PR is `secrets.*` in `if:` (GitHub rejects it)

Real-world signal arrives via real GitHub / GCP. Worth a separate ticket to harden the test scaffolding.

## Test plan

- [x] grep returns zero `secrets.X` references in `if:` lines
- [x] No behavioral change — same skip-when-not-configured semantics
- [ ] Real-fork smoke: user re-forks `tck517/agile-flow-gcp` after merge, repeats secrets/push, confirms deploy.yml runs and creates a Cloud Run service

## Coordination

After merge, user plans to nuke `tck517/agile-flow-gcp` and re-fork. The smoke test that surfaced this bug (and the previous bugs from today) becomes the acceptance signal for the entire chain: pre-flight checks, idempotent provisioning, retry helpers, ownership probe, org-policy override, WIF setup, attribute-condition fix, and now this. If `deploy.yml` runs end-to-end after the re-fork, today's eight (and now nine) PRs are validated together.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -28,16 +28,26 @@ jobs:
     steps:
       - name: Check for required secrets
         id: check_secrets
+        # GitHub Actions hard-rejects `secrets.*` in `if:` expressions, so
+        # we promote presence checks into outputs that downstream `if:`
+        # gates can read. `auth_method` selects between WIF (keyless,
+        # recommended) and service account key (workshop shortcut).
         run: |
           if [ -z "${{ secrets.GCP_PROJECT_ID }}" ]; then
             echo "skip=true" >> "$GITHUB_OUTPUT"
+            echo "auth_method=none" >> "$GITHUB_OUTPUT"
             echo "GCP_PROJECT_ID not configured — skipping deployment."
             echo "See docs/PLATFORM-GUIDE.md for setup instructions."
-          elif [ -z "${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}" ] && [ -z "${{ secrets.GCP_SA_KEY }}" ]; then
-            echo "skip=true" >> "$GITHUB_OUTPUT"
-            echo "Neither GCP_WORKLOAD_IDENTITY_PROVIDER nor GCP_SA_KEY configured — skipping."
-          else
+          elif [ -n "${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}" ]; then
             echo "skip=false" >> "$GITHUB_OUTPUT"
+            echo "auth_method=wif" >> "$GITHUB_OUTPUT"
+          elif [ -n "${{ secrets.GCP_SA_KEY }}" ]; then
+            echo "skip=false" >> "$GITHUB_OUTPUT"
+            echo "auth_method=sa-key" >> "$GITHUB_OUTPUT"
+          else
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+            echo "auth_method=none" >> "$GITHUB_OUTPUT"
+            echo "Neither GCP_WORKLOAD_IDENTITY_PROVIDER nor GCP_SA_KEY configured — skipping."
           fi
 
       - name: Checkout code
@@ -48,14 +58,14 @@ jobs:
       # Falls back to a service account JSON key if GCP_SA_KEY is set —
       # this is the workshop shortcut and is NOT recommended for production.
       - name: Authenticate to Google Cloud (Workload Identity Federation)
-        if: steps.check_secrets.outputs.skip != 'true' && secrets.GCP_WORKLOAD_IDENTITY_PROVIDER != ''
+        if: steps.check_secrets.outputs.auth_method == 'wif'
         uses: google-github-actions/auth@v2
         with:
           workload_identity_provider: ${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}
           service_account: ${{ secrets.GCP_SERVICE_ACCOUNT }}
 
       - name: Authenticate to Google Cloud (Service Account Key fallback)
-        if: steps.check_secrets.outputs.skip != 'true' && secrets.GCP_WORKLOAD_IDENTITY_PROVIDER == '' && secrets.GCP_SA_KEY != ''
+        if: steps.check_secrets.outputs.auth_method == 'sa-key'
         uses: google-github-actions/auth@v2
         with:
           credentials_json: ${{ secrets.GCP_SA_KEY }}
--- a/.github/workflows/preview-cleanup.yml
+++ b/.github/workflows/preview-cleanup.yml
@@ -21,10 +21,37 @@ jobs:
       PR_NUMBER: ${{ github.event.number }}
 
     steps:
+      - name: Check for required secrets
+        id: check_secrets
+        # GitHub Actions hard-rejects `secrets.*` in `if:` expressions, so
+        # we promote presence checks into outputs that downstream `if:`
+        # gates can read. `gcp_configured` gates GCP cleanup; `auth_method`
+        # selects between WIF and SA key; `neon_configured` gates the
+        # Neon branch deletion.
+        run: |
+          if [ -n "${{ secrets.GCP_PROJECT_ID }}" ]; then
+            echo "gcp_configured=true" >> "$GITHUB_OUTPUT"
+            if [ -n "${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}" ]; then
+              echo "auth_method=wif" >> "$GITHUB_OUTPUT"
+            elif [ -n "${{ secrets.GCP_SA_KEY }}" ]; then
+              echo "auth_method=sa-key" >> "$GITHUB_OUTPUT"
+            else
+              echo "auth_method=none" >> "$GITHUB_OUTPUT"
+            fi
+          else
+            echo "gcp_configured=false" >> "$GITHUB_OUTPUT"
+            echo "auth_method=none" >> "$GITHUB_OUTPUT"
+          fi
+          if [ -n "${{ secrets.NEON_API_KEY }}" ]; then
+            echo "neon_configured=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "neon_configured=false" >> "$GITHUB_OUTPUT"
+          fi
+
       # ── Neon: delete the PR branch database ───────────────────────────
 
       - name: Delete Neon branch
-        if: secrets.NEON_API_KEY != ''
+        if: steps.check_secrets.outputs.neon_configured == 'true'
         uses: neondatabase/delete-branch-action@v3
         continue-on-error: true
         with:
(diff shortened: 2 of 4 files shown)


va-reviewer:
## PR Review — #29

### Requirements

- [x] Zero `secrets.*` references remain in any `if:` expression — verified via `grep -nE "if:.*[^_]secrets\.[A-Z]" .github/workflows/*.yml` (zero matches)
- [x] Behavior preserved — same skip-when-not-configured semantics on all four workflows
- [x] `preview-cleanup.yml` gets a new `check_secrets` step (none existed before)
- [x] `rollback-production.yml` extends its existing step to emit `auth_method`; hard-fails if neither auth method is set (correct for a destructive admin action triggered via `workflow_dispatch`)
- [x] Each output encodes a single decision, not stuffed booleans (`skip`, `auth_method`, `gcp_configured`, `neon_configured`)
- [x] No new secret names; no behavioral drift

### Code Quality

- [x] Each workflow change is the **minimum** needed — no scope creep into unrelated cleanup
- [x] Comments at the top of every `check_secrets` step explain *why* the pattern exists (GitHub Actions hard-rejects `secrets.*` in `if:`). Future readers won't ask "why this shape?"
- [x] Bash style is internally consistent within each file (`-z` for skip-if-missing path, `-n` for set-output-if-present path)
- [x] No hardcoded values; all gating uses outputs from a single source-of-truth step

### Testing

- [x] No unit tests added — ticket explicitly excluded them; workflow YAML is validated by GitHub Actions on push, and there's no local linter for the secrets-in-if rule (yet)
- [x] **The PR validates itself in CI.** On this canonical repo, GCP secrets aren't configured. Before this PR, `Deploy PR Preview to Cloud Run` would have been `failure` (0s startup) due to the rejected `if:` expressions. After this PR, it's `skipping` — exactly what the new `check_secrets` step is supposed to produce on a repo without secrets. That outcome is the smoke test for the fix.
- [x] All standard CI checks pass (lint, typecheck, build, test, lint-agent-policies, python, node, Auto-fix, Trigger PR Review)

### Security

- [x] The PR **resolves** a GitHub-enforced security rule (secrets-in-if rejection prevents secret leakage through expression evaluation). The new pattern stores boolean presence flags, never the secret values themselves
- [x] `secrets.*` continues to appear (correctly) in `env:` and `with:` blocks, where GitHub permits it. The fix is precise about *where* the rule applies
- [x] No new command-injection vectors — bash heredocs use `${{ secrets.X }}` interpolation, which GitHub safely substitutes; `$GITHUB_OUTPUT` is written via `>>`, not via expression-evaluated assignment

### Recommendation

**GO**

P0 fix unblocking every workshop participant on day-1. The PR is the minimum-correct change: same semantics, no new abstractions, comments explain the GitHub Actions rule that caused this. The `Deploy PR Preview to Cloud Run` check transitioning from `failure` (before) to `skipping` (now) on a repo without GCP secrets is the smoke test — and it passed.

The "mocked tests verify shape, not acceptance" lesson is now confirmed three times in the same session (#26 quote-stripping, #27 attribute-condition, #29 secrets-in-if). The PR description rightly flags `actionlint` as the lever here — that's a real follow-up worth filing.

Ready to merge. After merge, user nukes the dry-run fork and re-forks to re-validate the entire pipeline end-to-end.

1. **`preview-deploy.yml` line 71 has a load-bearing-but-redundant-looking condition:** `steps.check_secrets.outputs.skip != 'true' && steps.check_secrets.outputs.neon_configured == 'true'`. Looks redundant because `skip == 'true'` could co-occur with `neon_configured == 'true'` (a fork with Neon but no GCP). The `&&` is correctly preventing the Neon-branch step from running when the rest of the deploy is skipped. A one-line comment naming the load-bearing nature would help future readers not "simplify" it away.

2. **`preview-cleanup.yml` doesn't define a unified `skip` output**, unlike the other three. That's correct — cleanup is *independent* per resource (Neon and GCP can be cleaned separately), not all-or-nothing. Worth a one-line code comment explaining this design choice; today it's implicit.

3. **Bash `if` style differs across workflows:** flat 3-branch (`if -z … elif -n … elif -n … else`) in deploy/preview-deploy vs. nested (`if -n … then if … fi else …`) in preview-cleanup. Both correct; today's mix is defensible because cleanup has the additional `gcp_configured` to gate on. Picking one style across all four would be a small future-maintenance win. Out of scope here.

4. **Add `actionlint` to CI / pre-push** as a separate ticket. It catches `secrets.*` in `if:` (and many other workflow gotchas) before push. Would have prevented this exact bug class from reaching `main`. The PR description already raises this — worth filing.

5. **The `Run Alembic migrations` step gates on `secrets.PRODUCTION_DATABASE_URL` via `env:` (correct, allowed) and an inline `[ -z "$DATABASE_URL" ]` check.** Different shape from the auth-step pattern. Not a regression — same as before this PR. If we ever do a "clean up gating patterns across workflows" pass, this one's worth folding into the same approach.

---

PR #31, opened 2026-04-28
fix(scripts+ci): grant SA token-creator + pin Cloud Run runtime SA to deployer

### LINKED ISSUE #30: fix(scripts+ci): grant SA token-creator + pin Cloud Run runtime SA to deployer
## Problem Statement

Two related defects discovered live during workshop smoke 2026-04-28. Both block the deploy workflow's first run on a fresh fork; bundled because they share the same test scaffolding and both unblock the same end-to-end smoke.

### Defect 1 — Step 5.5c missing `serviceAccountTokenCreator`

`provision-gcp-project.sh` Step 5.5c (added in #5) grants `roles/iam.workloadIdentityUser` on the deployer SA to the workshop participant's GitHub repo principal set. **That role alone is insufficient** for the workflow's actual use case.

`google-github-actions/auth@v2` in this stack runs in **impersonation mode** because the workflow passes both `workload_identity_provider` and `service_account`. In that mode the action:

1. Uses the federated identity (granted by `workloadIdentityUser`) to authenticate to GCP
2. Then calls `iam.serviceAccountTokenCreator.generateAccessToken()` to mint a real OAuth access token *for the deployer SA*
3. That access token is what `gcloud auth configure-docker`, `docker push`, and the Cloud Run deploy step actually use

Step 2 requires **`roles/iam.serviceAccountTokenCreator`** on the SA itself, granted to the same federated principal. Without it, the federated identity can authenticate but cannot mint tokens, and the next step that needs SA credentials fails.

Live failure during smoke 2026-04-28 on `af-worker-2026-05a`:

```text
ERROR: (gcloud.auth.docker-helper) There was a problem refreshing your current
auth tokens: ('Unable to acquire impersonated credentials',
  '{"error": {"code": 403,
    "message": "Permission \\'iam.serviceAccounts.getAccessToken\\' denied
                on resource (or it may not exist).",
    "details": [{"reason": "IAM_PERMISSION_DENIED",
                 "metadata": {"permission": "iam.serviceAccounts.getAccessToken"}}]}}'
)
...
denied: Unauthenticated request. Unauthenticated requests do not have permission
"artifactregistry.repositories.uploadArtifacts" on resource ".../agile-flow"
```

The `iam.serviceAccounts.getAccessToken` permission is exactly what `roles/iam.serviceAccountTokenCreator` grants. The cascade — `gcloud auth configure-docker` couldn't refresh tokens → docker push fell back to anonymous → Artifact Registry rejected as unauthenticated — all unwinds from the missing role.

### Defect 2 — `deploy.yml` doesn't pin the Cloud Run runtime service account

After the token-creator binding was applied manually and the workflow was re-run, build + push succeeded. The deploy step then failed:

```text
spec.template.spec.containers[0].env[1].value_from.secret_key_ref.name:
Permission denied on secret: projects/.../secrets/database-url/versions/latest
for Revision service account 715215775688-compute@developer.gserviceaccount.com.
The service account used must be granted the 'Secret Manager Secret Accessor' role.
```

The deploy job *deploys* as the deployer SA (which already has `roles/secretmanager.secretAccessor` on the project from Step 5). But the **Cloud Run revision runs** under the *legacy Compute Engine default SA* (`<num>-compute@developer.gserviceaccount.com`) by default, and that SA does NOT have the role. Modern Cloud Run best practice is to pin a specific runtime SA — and the deployer SA already has every runtime role we need.

The fix is to pass `--service-account="$GCP_SERVICE_ACCOUNT"` to the `gcloud run deploy` invocation in `deploy.yml`, so the revision runs as the same SA that deployed it. No new bindings needed.

### Pattern

This is the **fourth** workshop-blocker in the same provisioning flow this week, all bugs in roles or rules `provision-gcp-project.sh` doesn't yet handle. Pattern: the script's IAM model was hand-derived from the platform-guide narrative, which doesn't enumerate every role `google-github-actions/auth` needs in impersonation mode, and doesn't think clearly about the deployer-SA-vs-runtime-SA distinction.

**Effort Estimate:** XS
**Priority:** P0 (workshop dry-run is 2026-05-01, 3 days from filing; current behavior breaks every fork's first deploy push)

---

## A. Environment Context

- Stack: Bash, same as #5
- Files to modify:
  - MODIFY `scripts/provision-gcp-project.sh` Step 5.5c — add a second `add-iam-policy-binding` for `roles/iam.serviceAccountTokenCreator` on the deployer SA, with the same `principalSet://...attribute.repository/${GITHUB_USERNAME}/${WIF_REPO}` member
  - MODIFY `.github/workflows/deploy.yml` and `.github/workflows/preview-deploy.yml` — pass `service_account: ${{ env.GCP_SERVICE_ACCOUNT }}` to the `google-github-actions/deploy-cloudrun@v2` action, so the Cloud Run revision runs as the deployer SA (which already has `roles/secretmanager.secretAccessor`)
  - MODIFY `scripts/provision-gcp-project.test.sh` — extend the existing test 13 (and/or add a focused test) to assert *both* role bindings are issued
  - MODIFY `docs/PLATFORM-GUIDE.md` Step 5 manual-fallback snippet — same one-line addition so copy-paste users hit the same fix
- Reference:
  - `google-github-actions/auth` README on impersonation: <https://github.com/google-github-actions/auth#indirect-wif>
  - GCP IAM doc on `roles/iam.serviceAccountTokenCreator`: <https://cloud.google.com/iam/docs/service-account-permissions#token-creator-role>
  - Cloud Run docs on per-service runtime SA: <https://cloud.google.com/run/docs/configuring/services/service-identity>

---

## B. Guardrails

- The new binding MUST be on the **deployer SA** (`deployer@<project>.iam.gserviceaccount.com`), not at project level. Token-creator at project level would over-broaden — the SA-scoped binding is exactly what `auth@v2` needs.
- The new binding MUST use the **same `principalSet://` member** as the existing `workloadIdentityUser` binding (`attribute.repository/${GITHUB_USERNAME}/${WIF_REPO}`). The two roles together describe one trust relationship; the principals must match.
- Both bindings MUST be idempotent. `add-iam-policy-binding` is at gcloud's level; re-running the script must not error.
- Do NOT consolidate into a custom role. Workshop scope; predefined roles are clearer for facilitators reading the audit log.
- Do NOT add the role at the project level via Step 5's IAM-roles loop. That loop is for the deployer SA's *outbound* permissions (Cloud Run admin, Artifact Registry writer, etc.) — different concern from "who can impersonate this SA."

---

## C. Happy Path

1. Facilitator runs `BILLING_ACCOUNT_ID=… ./scripts/workshop-setup.sh roster.csv` for the first time on a fresh project (or re-runs after a failed deploy).
2. Step 5.5c logs:
   - `[bind] roles/iam.workloadIdentityUser <- ${GITHUB_USERNAME}/${WIF_REPO}` (existing)
   - `[bind] roles/iam.serviceAccountTokenCreator <- ${GITHUB_USERNAME}/${WIF_REPO}` (new)
3. Participant pushes to `main` on their fork.
4. `deploy.yml` runs the WIF auth step → succeeds. `gcloud auth configure-docker` mints an SA access token via the new role → succeeds. `docker push` reaches Artifact Registry as the deployer SA → succeeds. Cloud Run deploy lands.
5. No `IAM_PERMISSION_DENIED` on `iam.serviceAccounts.getAccessToken` anywhere in the workflow log.

---

### Defect 1 — provisioner

- [ ] `scripts/provision-gcp-project.sh` Step 5.5c issues two `add-iam-policy-binding` calls — one for `workloadIdentityUser`, one for `serviceAccountTokenCreator` — both with the same `principalSet://...attribute.repository/<user>/<repo>` member
- [ ] Both bindings are gated on `GITHUB_USERNAME` being set (same condition as the rest of Step 5.5)
- [ ] Both bindings are idempotent — re-running the script with the bindings already in place is a no-op
- [ ] `scripts/provision-gcp-project.test.sh`: existing test 13 (or a new test 15) asserts that *both* role bindings appear in the gcloud call log when `GITHUB_USERNAME` is set
- [ ] `docs/PLATFORM-GUIDE.md` Step 5 manual-fallback snippet shows both bindings

### Defect 2 — deploy workflows

- [ ] `.github/workflows/deploy.yml` `Deploy to Cloud Run` step passes `service_account: ${{ env.GCP_SERVICE_ACCOUNT }}` (or equivalent flag) to `google-github-actions/deploy-cloudrun@v2`, so the revision runs as the deployer SA
- [ ] `.github/workflows/preview-deploy.yml` makes the same change to its corresponding deploy step
- [ ] No participant-side change is required — facilitator emails remain the four-secret list

### Both

- [ ] All 38 helper-test assertions still pass (no regression)
- [ ] Real-GCP smoke deferred to user's pending re-run on `af-worker-2026-05a` (which is what surfaced both bugs)

---

## Why P0

This is a hard block on the workshop's "deploy from your fork" step. Every participant's first push will fail at docker push with the same generic permission error, which gives the facilitator no clear signal pointing at the missing role. Three days to dry-run; the dry-run cannot reach a deployed Cloud Run service without this fix.

---

## Note on the recurring pattern

This is the fourth workshop-blocking bug discovered the same way:
- #14 API-enable retry (eventual consistency)
- #16 SA-binding retry (eventual consistency)
- #19 org-policy override (Cloud Identity Free default)
- **This ticket** (impersonation requires two roles, not one)

Each was a hidden requirement in GCP's IAM model that wasn't visible from script-level reasoning. The next preventive step (out of scope here) is adding `actionlint` and a "first-deploy validation" smoke test that runs against a throwaway project before the workshop. Worth filing once the immediate workshop is unblocked.

va-worker:
## Summary

Closes #30. **P0 — workshop dry-run is in 3 days.**

Two related defects discovered live during workshop smoke today, bundled because they share test scaffolding and both unblock the same end-to-end deploy.

### Defect 1: Step 5.5c missing serviceAccountTokenCreator

`google-github-actions/auth@v2` in impersonation mode needs **two** roles on the deployer SA. PR #5 granted only `roles/iam.workloadIdentityUser`. The second role, `roles/iam.serviceAccountTokenCreator`, is what `gcloud auth configure-docker` actually uses to refresh tokens for `docker push`.

Without it: `iam.serviceAccounts.getAccessToken denied` → docker push falls back to anonymous → Artifact Registry rejects → workflow fails.

Fix: loop over both roles in Step 5.5c. `add-iam-policy-binding` is idempotent at gcloud's level.

### Defect 2: deploy workflows don't pin Cloud Run runtime SA

After the token-creator fix landed manually and the workflow re-ran, `docker push` succeeded — and the deploy step then failed with:

```text
Permission denied on secret database-url for Revision service account
<num>-compute@developer.gserviceaccount.com.
```

Cloud Run uses the legacy Compute Engine default SA at runtime by default. That SA has none of the roles provisioned in Step 5 (`secretmanager.secretAccessor`, etc.). The **deployer SA** does have them.

Fix: pass the deployer SA as the runtime SA. One service account for deploy + runtime → clearer audit trail, no need to grant roles to a default SA we didn't create.

- `deploy.yml`: `service_account: ${{ secrets.GCP_SERVICE_ACCOUNT }}` on `deploy-cloudrun@v2`
- `preview-deploy.yml`: `--service-account="${{ secrets.GCP_SERVICE_ACCOUNT }}"` on raw `gcloud run deploy`

## Tests

`scripts/provision-gcp-project.test.sh` test 13 now asserts **both** `[bind]` log lines. The new assertions:

```bash
grep -q "\[bind\] roles/iam.workloadIdentityUser <- alice-gh/agile-flow-gcp" "$T13/stdout.log"
grep -q "\[bind\] roles/iam.serviceAccountTokenCreator <- alice-gh/agile-flow-gcp" "$T13/stdout.log"
```

40/40 assertions pass (existing 38 + 2 new). No regression.

## Implementation note

Step 5.5c now uses a `for wif_role in ...; do` loop over the two roles instead of two separate `gcloud` invocations. Same idempotency, smaller diff, easier to add a third role in the future if it turns out to also be needed (it almost certainly won't, but the pattern is now established).

## Acceptance

User has confirmed the smoke-test philosophy: clean end-to-end run from a freshly-nuked starting point, not patches against in-flight projects. After this merges, user nukes `tck517/agile-flow-gcp` and re-forks. If the deploy completes cleanly through `Show deployed URL`, today's chain (PRs #14, #16, #19, #22, #5, #27, #29, #30) is end-to-end validated for the workshop.

## Test plan

- [x] 40/40 helper tests pass
- [x] No regression on wrapper tests
- [x] markdownlint clean
- [ ] Real-GCP smoke: user re-forks after merge

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -117,6 +117,12 @@ jobs:
           service: ${{ env.SERVICE_NAME }}
           region: ${{ env.GCP_REGION }}
           image: ${{ steps.build.outputs.image }}
+          # Pin the runtime SA to the deployer SA. By default Cloud Run
+          # uses the legacy Compute Engine default SA at runtime, which
+          # has none of the roles we provisioned (e.g. secretAccessor).
+          # Using one SA for deploy + runtime keeps the audit trail clear
+          # and avoids granting roles to a default SA we did not create.
+          service_account: ${{ secrets.GCP_SERVICE_ACCOUNT }}
           flags: |
             --port=8080
             --memory=512Mi
--- a/.github/workflows/preview-deploy.yml
+++ b/.github/workflows/preview-deploy.yml
@@ -142,9 +142,15 @@ jobs:
             NEON_URL_FLAG="--update-env-vars=DATABASE_URL=${{ steps.neon.outputs.db_url_pooled }}"
           fi
 
+          # --service-account pins the runtime SA to the deployer SA so
+          # the revision can read mounted Secret Manager secrets and any
+          # other resource the deployer SA already has access to. Without
+          # this, Cloud Run uses the legacy Compute Engine default SA at
+          # runtime, which has none of the roles provisioned in Step 5.
           gcloud run deploy "${{ env.SERVICE_NAME }}" \
             --image="${{ steps.build.outputs.image }}" \
             --region="${{ env.GCP_REGION }}" \
+            --service-account="${{ secrets.GCP_SERVICE_ACCOUNT }}" \
             --tag="pr-${{ env.PR_NUMBER }}" \
             --no-traffic \
             --port=8080 \
--- a/docs/PLATFORM-GUIDE.md
+++ b/docs/PLATFORM-GUIDE.md
@@ -206,12 +206,16 @@ gcloud iam workload-identity-pools providers create-oidc github \
 # Get the project number (different from project ID)
 PROJECT_NUMBER=$(gcloud projects describe YOUR_PROJECT_ID --format='value(projectNumber)')
 
-# Allow the GitHub repo to impersonate the deployer service account
-gcloud iam service-accounts add-iam-policy-binding \
-  "deployer@YOUR_PROJECT_ID.iam.gserviceaccount.com" \
-  --role="roles/iam.workloadIdentityUser" \
-  --member="principalSet://iam.googleapis.com/projects/${PROJECT_NUMBER}/locations/global/workloadIdentityPools/github/attribute.repository/GITHUB_USER/REPO_NAME" \
-  --project=YOUR_PROJECT_ID
+# Allow the GitHub repo to impersonate the deployer service account.
+# Two roles are required: workloadIdentityUser to authenticate as the SA,
+# and serviceAccountTokenCreator to mint access tokens (gcloud, docker push).
+for role in roles/iam.workloadIdentityUser roles/iam.serviceAccountTokenCreator; do
+  gcloud iam service-accounts add-iam-policy-binding \
+    "deployer@YOUR_PROJECT_ID.iam.gserviceaccount.com" \
+    --role="$role" \
+    --member="principalSet://iam.googleapis.com/projects/${PROJECT_NUMBER}/locations/global/workloadIdentityPools/github/attribute.repository/GITHUB_USER/REPO_NAME" \
+    --project=YOUR_PROJECT_ID
+done
 ```
 
 The WIF provider resource name to paste into GitHub secrets is:
--- a/scripts/provision-gcp-project.sh
+++ b/scripts/provision-gcp-project.sh
@@ -350,15 +350,30 @@ if [[ -n "${GITHUB_USERNAME:-}" ]]; then
   fi
 
   # 5.5c: Bind the deployer SA so the specific repo can impersonate it.
+  # google-github-actions/auth@v2 in impersonation mode (which our deploy
+  # workflow uses by passing both `workload_identity_provider` and
+  # `service_account`) needs TWO roles:
+  #
+  #   roles/iam.workloadIdentityUser    — lets the federated identity
+  #                                        authenticate AS the SA via WIF
+  #   roles/iam.serviceAccountTokenCreator — lets the federated identity
+  #                                        MINT access tokens for the SA
+  #                                        (gcloud auth, docker push, etc.)
+  #
+  # Granting only the first leaves the deploy step's docker-push call
+  # failing with `iam.serviceAccounts.getAccessToken denied`.
+  #
   # add-iam-policy-binding is idempotent — re-running with the same member
   # is a no-op.
   WIF_MEMBER="principalSet://iam.googleapis.com/projects/${PROJECT_NUMBER}/locations/global/workloadIdentityPools/${WIF_POOL}/attribute.repository/${GITHUB_USERNAME}/${WIF_REPO}"
-  echo "[bind] roles/iam.workloadIdentityUser <- ${GITHUB_USERNAME}/${WIF_REPO}"
-  gcloud iam service-accounts add-iam-policy-binding "$SA_EMAIL" \
-    --role="roles/iam.workloadIdentityUser" \
-    --member="$WIF_MEMBER" \
-    --project="$GCP_PROJECT_ID" \
-    --quiet >/dev/null
+  for wif_role in roles/iam.workloadIdentityUser roles/iam.serviceAccountTokenCreator; do
+    echo "[bind] $wif_role <- ${GITHUB_USERNAME}/${WIF_REPO}"
+    gcloud iam service-accounts add-iam-policy-binding "$SA_EMAIL" \
+      --role="$wif_role" \
+      --member="$WIF_MEMBER" \
+      --project="$GCP_PROJECT_ID" \
+      --quiet >/dev/null
+  done
 
   WIF_PROVIDER_RESOURCE="projects/${PROJECT_NUMBER}/locations/global/workloadIdentityPools/${WIF_POOL}/providers/${WIF_PROVIDER}"
 else
--- a/scripts/provision-gcp-project.test.sh
+++ b/scripts/provision-gcp-project.test.sh
@@ -640,6 +640,24 @@ else
   FAIL=$((FAIL + 1))
 fi
 
+# Both WIF roles must be bound. workloadIdentityUser alone leaves the deploy
+# step unable to mint access tokens for docker push. We learned that the
+# hard way during smoke 2026-04-28.
+if grep -q "\[bind\] roles/iam.workloadIdentityUser <- alice-gh/agile-flow-gcp" "$T13/stdout.log"; then
+  echo -e "  ${GREEN}✓${NC} workloadIdentityUser binding logged"
+  PASS=$((PASS + 1))
+else
+  echo -e "  ${RED}✗${NC} expected '[bind] roles/iam.workloadIdentityUser <- alice-gh/agile-flow-gcp'"
+  FAIL=$((FAIL + 1))
+fi
+if grep -q "\[bind\] roles/iam.serviceAccountTokenCreator <- alice-gh/agile-flow-gcp" "$T13/stdout.log"; then
+  echo -e "  ${GREEN}✓${NC} serviceAccountTokenCreator binding logged"
+  PASS=$((PASS + 1))
+else
+  echo -e "  ${RED}✗${NC} expected '[bind] roles/iam.serviceAccountTokenCreator <- alice-gh/agile-flow-gcp'"
+  FAIL=$((FAIL + 1))
+fi
+
 # Final summary line should print the WIF provider resource path with project number 12345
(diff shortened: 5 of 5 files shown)


va-reviewer:
## PR Review — #31

### Requirements

#### Defect 1 — provisioner
- [x] Step 5.5c issues both bindings via a `for wif_role in ...; do` loop
- [x] Both gated on `GITHUB_USERNAME` being set (same outer `if`)
- [x] Both idempotent at gcloud's level
- [x] Test 13 asserts both `[bind]` log lines explicitly

#### Defect 2 — deploy workflows
- [x] `deploy.yml` passes `service_account: ${{ secrets.GCP_SERVICE_ACCOUNT }}` to `deploy-cloudrun@v2`
- [x] `preview-deploy.yml` adds `--service-account="${{ secrets.GCP_SERVICE_ACCOUNT }}"` to raw `gcloud run deploy`
- [x] Inline comments on both name the runtime-vs-deployer-SA distinction
- [x] PLATFORM-GUIDE manual snippet uses the same loop shape as the script

#### Both
- [x] 40/40 helper tests pass
- [x] `shellcheck`, `bash -n`, and `markdownlint` clean
- [x] Real-GCP smoke deferred per user's philosophy (clean fresh-fork run is the acceptance signal)

### Code Quality

- [x] Each of the 5 file changes touches exactly the right place — minimal scope
- [x] **The Step 5.5c docstring is the gold standard** — names both roles, what each one does, and the exact failure signature of granting only the first. This prevents the regression from coming back in 6 months.
- [x] Workflow comments name *why* the runtime SA needs pinning (legacy Compute Engine default SA has none of the provisioned roles)
- [x] No hardcoded values; all auth flows through the existing `GCP_SERVICE_ACCOUNT` secret

### Testing

- **Test 13's two new assertions** lock in the load-bearing invariant: both `[bind] roles/...` log lines appear when `GITHUB_USERNAME` is set. Either role being silently dropped fails the test loudly with the expected pattern shown in the failure message.
- 40/40 helper tests pass (existing 38 + 2 new)
- All CI checks green; preview-deploy on the canonical repo correctly skips (the new check_secrets pattern from #29 is doing its job)

### Security

- [x] **`roles/iam.serviceAccountTokenCreator` correctly scoped to the SA itself** (`add-iam-policy-binding "$SA_EMAIL"`), not project-level. The federated identity can only mint tokens for *this* SA, not arbitrary SAs.
- [x] **Pinning the runtime SA away from the Compute Engine default is a security improvement** — default SAs are historically over-privileged on legacy projects.
- [x] No new command-injection vectors — `${{ secrets.X }}` interpolation in `with:` and inline gcloud is GitHub's safe substitution

### Recommendation

**GO**

P0 unblocking the second-to-last bug between "scripts work locally" and "fresh fork runs end-to-end." The runtime-SA pin is the correct architectural choice (modern Cloud Run best practice). Tests catch the load-bearing invariant. The Step 5.5c docstring is the gold standard for preventing this regression from coming back. Ready to merge.

This PR closes the IAM gap that's been chasing the workshop all afternoon. After merge + fresh-fork smoke: today's chain (PRs #5, #14, #16, #19, #22, #27, #29, #31) is end-to-end validated.

1. **`for wif_role in a b; do` matches the inline-list shape, but Step 5's existing deployer-SA-roles loop uses an array** (`ROLES=(...)` + `for role in "${ROLES[@]}"`). Both correct; an array would match precedent. Worth aligning if anyone touches both blocks in the same PR later.

2. **`GCP_SERVICE_ACCOUNT` could be hoisted to the job's `env:` block** in `deploy.yml` for symmetry with `GCP_PROJECT_ID` (which is hoisted today). Stylistic — direct `${{ secrets.X }}` in `with:` is also valid.

3. **No YAML-shape test for the `service_account:` line.** Same gap as #29's `--attribute-condition`. When you file the actionlint follow-up, consider also adding a `grep -q "service_account: " .github/workflows/deploy.yml` shape check so removing this line would fail a local check before CI.

4. **The Step 5.5c docstring could mention the role is scoped to *this SA*, not project-level** — explicit least-privilege framing for security-conscious future readers. The implementation is already correct; just the comment doesn't say it.

5. **The runtime-SA pin works on first deploy** but doesn't *retroactively* migrate revisions that were previously deployed without it. Cloud Run's default traffic-split moves to the latest revision, so this isn't a practical problem — but worth noting in the runbook for facilitators who hit a half-deployed state.

---

Issue #33, opened 2026-04-28
feat(scripts): provision Neon branch + database-url secret per attendee


va-worker:
## Problem Statement

Today the workshop's Neon database story has two gaps:

1. **No automated `database-url` Secret Manager secret per project.** Every facilitator/participant has to manually run `echo url | gcloud secrets create database-url ...` per participant project. With 8 participants that's 8 hand-typed gcloud invocations.

2. **No isolation between participants' database state.** A shared Neon database means alice's `DROP TABLE` would affect bob's working state. Schema migrations from one participant's experiments would interfere with everyone else's. The current "one shared connection string for the cohort" model only works if every participant runs identical code at identical times — the opposite of a workshop where everyone builds something different.

This ticket replaces the original "facilitator provides a single `NEON_DATABASE_URL`" design (filed 2026-04-28) with a **branch-per-attendee** model that closes both gaps.

**Design summary:** one Neon project per cohort. Provisioner calls the Neon API to create one branch per participant, reads the resulting pooled connection string, and writes it to that participant's `database-url` Secret Manager secret. Each participant's branch is an isolated copy-on-write Postgres instance — alice's schema changes don't touch bob's. Per-PR previews already work via `preview-deploy.yml`'s `neondatabase/create-branch-action`; the only template-side change is to set the `parent:` to the attendee's branch (handled by separate ticket; this one is provisioner-only).

**Parent Epic:** vibeacademy/agile-flow-gcp#2 — Workshop-ready hardening of the GCP fork
**Effort Estimate:** S
**Priority:** P0 (workshop dry-run is 2026-05-01, 3 days from filing; current behavior fails the deploy step at the secret-mount step)

---

## A. Environment Context

- Stack: Bash, gcloud, Neon REST API
- Required env vars (facilitator-supplied at wrapper invocation):
  - `NEON_API_KEY` — used to create branches
  - `NEON_PROJECT_ID` — the cohort's Neon project
  - `NEON_PARENT_BRANCH_ID` (optional) — defaults to the project's `main` branch ID
- Files to modify:
  - MODIFY `scripts/provision-gcp-project.sh` — add a Step 5.7 (or wherever fits) that creates the Neon branch and the `database-url` Secret Manager secret. Idempotent: if a branch named `${NEON_BRANCH_NAME}` already exists, fetch its connection string instead of failing
[...]

PR #34, opened 2026-04-28
fix(ci): pass --service-account via flags, not as top-level input

### LINKED ISSUE #32: fix(ci): deploy-cloudrun service_account must be passed via flags, not as a top-level input
## Problem Statement

PR #31 added `service_account: ${{ secrets.GCP_SERVICE_ACCOUNT }}` as a top-level input to the `google-github-actions/deploy-cloudrun@v2` action in `deploy.yml`. **The action does not accept `service_account` as a top-level input** — it only accepts a fixed list (`service`, `job`, `metadata`, `image`, `source`, `suffix`, `env_vars`, `env_vars_file`, `env_vars_update_strategy`, `secrets`, `secrets_update_strategy`, `labels`, `skip_default_labels`, `tag`, `timeout`, `flags`, `no_traffic`, `wait`, `revision_traffic`, `tag_traffic`, `update_traffic_flags`, `project_id`, `region`, `gcloud_version`, `gcloud_component`).

GitHub Actions logs the unknown input as a **warning** (not an error) and silently drops it. The deploy still runs — but without the runtime SA pin. So the original bug from #30 (defect 2) is back: the Cloud Run revision uses the legacy Compute Engine default SA at runtime instead of the deployer SA.

Live reproducer 2026-04-28 on the post-#31 main:

```text
Warning: Unexpected input(s) 'service_account', valid inputs are
['service', 'job', 'metadata', 'image', 'source', 'suffix', 'env_vars', ...,
'flags', ...]
Run google-github-actions/deploy-cloudrun@v2
Successfully authenticated
Running: gcloud run deploy agile-flow-app --image ... --update-env-vars ... --update-secrets ...
[no --service-account flag in the running gcloud command]
```

The fix is to pass `--service-account=${{ secrets.GCP_SERVICE_ACCOUNT }}` via the action's existing `flags:` input, which forwards arbitrary flags to `gcloud run deploy`. `preview-deploy.yml` already uses raw `gcloud run deploy` directly and got this right in #31 — it's only the action-based call site that needs changing.

**Effort Estimate:** XS (one-line fix)
**Priority:** P0 (workshop dry-run is in 3 days; this re-opens the runtime-SA bug from #30)

---

## A. Environment Context

- Stack: GitHub Actions YAML
- Files to modify:
  - MODIFY `.github/workflows/deploy.yml` — move `service_account:` from top-level input to a `--service-account=...` line in `flags:`
- Reference:
  - `google-github-actions/deploy-cloudrun@v2` README: <https://github.com/google-github-actions/deploy-cloudrun#inputs>
  - The action's input list is shown in the live error log above

---

## B. Guardrails

- Do NOT change `preview-deploy.yml`. It uses raw `gcloud run deploy --service-account=...` and that's correct.
- The test scaffolding in `provision-gcp-project.test.sh` doesn't cover the workflow YAML directly. No new tests required for this fix; same gap as #29's secrets-in-if (would require `actionlint` or a real workflow run to catch).
- Do NOT pass any other flags via the action's top-level inputs that aren't in the official list. Use `flags:` for anything not in: `service`, `image`, `region`, `env_vars`, `secrets`, `tag`, `no_traffic`, `flags`, etc.

---

## C. Happy Path

1. PR merges. `deploy.yml`'s `service_account:` input is replaced with a `--service-account=...` line inside `flags:`.
2. User re-pushes a trivial commit on their fork.
3. Deploy step runs `gcloud run deploy agile-flow-app --image ... --service-account=deployer@... ...`. No "Unexpected input" warning.
4. Cloud Run revision runs as the deployer SA, which has `roles/secretmanager.secretAccessor` from Step 5.
5. The `database-url` secret mount succeeds (assuming the secret exists — that's a separate ticket).

---

## D. Definition of Done

- [ ] `.github/workflows/deploy.yml` removes the top-level `service_account:` input from the `deploy-cloudrun` step
- [ ] `flags:` block gains `--service-account=${{ secrets.GCP_SERVICE_ACCOUNT }}`
- [ ] No "Unexpected input(s) 'service_account'" warning in CI runs after merge
- [ ] No regression on existing helper tests (40/40 still pass)
- [ ] markdownlint clean
- [ ] Real-GCP smoke deferred to user's next nuke-and-fork run

---

## Note

This is a real bug from #31's review that I missed. The `service_account` field is a documented input for the **`auth@v2`** action (where we use it correctly) but NOT for the `deploy-cloudrun@v2` action. Easy to confuse since both are in the `google-github-actions` org. The action's README does not advertise this asymmetry; only the live runtime warning surfaces it.

va-worker:
## Summary

Closes #32. **P0 — undoes a silent regression I introduced in #31.**

PR #31 added `service_account: ${{ secrets.GCP_SERVICE_ACCOUNT }}` as a top-level input to `deploy-cloudrun@v2`. **The action does not accept `service_account` as a top-level input.** GitHub Actions logged a warning and silently dropped the input, so the runtime-SA pin from #30 (defect 2) silently never happened post-merge.

Live evidence:

```text
Warning: Unexpected input(s) 'service_account', valid inputs are
['service', 'job', 'metadata', 'image', 'source', 'suffix',
 'env_vars', ..., 'flags', ...]
```

## Fix

The action's `flags:` input forwards arbitrary args to `gcloud run deploy`. Move `--service-account=...` there:

```yaml
flags: |
  --service-account=${{ secrets.GCP_SERVICE_ACCOUNT }}
  --port=8080
  ...
```

`preview-deploy.yml` is unaffected — it uses raw `gcloud run deploy --service-account=…`, which has been correct since #31.

## The asymmetry that bit me

`auth@v2` *does* accept `service_account:` as a top-level input. `deploy-cloudrun@v2` does not. Same `google-github-actions/*` org, same parameter name, different APIs. Easy to confuse. The new comment in `deploy.yml` names this trap explicitly so the next reader doesn't repeat the mistake.

## Test plan

- [x] secrets-in-if invariant still holds: `grep -nE "if:.*[^_]secrets\.[A-Z]"` returns zero matches
- [x] markdownlint clean (no doc changes)
- [ ] CI green
- [ ] Real-GCP smoke at user's next nuke-and-fork — the runtime-SA pin should actually take effect this time

## Why a separate PR (not amended into #31)

#31 is merged. The fix has to land on top of main. One-line PR is the smallest unit; bundling with the database-url-secret automation (filed as #33, P1) would conflate two scopes.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -117,13 +117,14 @@ jobs:
           service: ${{ env.SERVICE_NAME }}
           region: ${{ env.GCP_REGION }}
           image: ${{ steps.build.outputs.image }}
-          # Pin the runtime SA to the deployer SA. By default Cloud Run
-          # uses the legacy Compute Engine default SA at runtime, which
-          # has none of the roles we provisioned (e.g. secretAccessor).
-          # Using one SA for deploy + runtime keeps the audit trail clear
-          # and avoids granting roles to a default SA we did not create.
-          service_account: ${{ secrets.GCP_SERVICE_ACCOUNT }}
+          # Runtime-SA pin: --service-account is passed via flags: rather
+          # than as a top-level input, because deploy-cloudrun@v2 does NOT
+          # accept service_account as a top-level input (only auth@v2
+          # does — the asymmetry is easy to miss). Without this flag,
+          # Cloud Run uses the legacy Compute Engine default SA at runtime,
+          # which has none of the roles we provisioned (e.g. secretAccessor).
           flags: |
+            --service-account=${{ secrets.GCP_SERVICE_ACCOUNT }}
             --port=8080
             --memory=512Mi
             --cpu=1


va-reviewer:
## PR Review — #34

### Requirements

- [x] `service_account:` removed from top-level `deploy-cloudrun@v2` inputs
- [x] `--service-account=${{ secrets.GCP_SERVICE_ACCOUNT }}` added to `flags:` block
- [x] No "Unexpected input(s) 'service_account'" warning expected on next run
- [x] `preview-deploy.yml` unchanged (correctly uses raw `gcloud run deploy --service-account=…` already)
- [x] No regression on existing workflow behavior — the deploy step still uses `deploy-cloudrun@v2`, just with a valid input shape

### Code Quality

- [x] Comment **correctly names the asymmetry** between `auth@v2` (accepts `service_account:`) and `deploy-cloudrun@v2` (does not). Strong craft — prevents the same mistake from being made by the next reader.
- [x] `--service-account` placed first in the `flags:` list — visually prominent, easy to spot when reading
- [x] Single-line value interpolation works correctly inside `flags: |` (each line = one argument; no shell quoting needed)
- [x] No new abstractions, no scope creep

### Testing

- [x] No unit tests added — same workflow-YAML-validation gap as #29 and #31; `actionlint` follow-up still pending
- [x] All standard CI checks pass (lint, typecheck, build, test, lint-agent-policies, python, node, Auto-fix, Trigger PR Review)
- [x] `Deploy PR Preview to Cloud Run` correctly skips on the canonical repo (no GCP secrets configured)
- [x] **The trigger for this PR was the smoke test itself.** The runtime warning told us exactly what to do; this PR does that. Re-smoking after merge is the acceptance signal — the warning disappears.

### Security

- [x] No new secrets, tokens, or credentials
- [x] **This PR re-enables** the runtime-SA pin that #31 silently broke — a security improvement (Cloud Run revision runs as the deployer SA, not the legacy Compute Engine default)
- [x] No injection vectors — `${{ secrets.X }}` in `flags:` is GitHub's safe substitution
- [x] secrets-in-if invariant from #29 holds (no `secrets.X` references in `if:` lines)

### Recommendation

**GO**

Small, high-confidence fix. PR #31 silently regressed the runtime-SA pin because `deploy-cloudrun@v2` doesn't accept `service_account` as a top-level input — GitHub Actions logged a warning and dropped it. Moving the value into `flags:` is the documented and correct way. The new comment names the asymmetry explicitly so the trap doesn't catch the next reader.

Self-validating: the next workflow run on a fresh fork will either show the "Unexpected input" warning gone (fix worked) or fail loudly (fix didn't take). Either is a clear signal.

Ready to merge.

1. **`actionlint` would have caught this at PR time** — same lesson as #29's secrets-in-if and #27's missing `--attribute-condition`. The PR description rightly raises it. Worth filing as a follow-up: add `actionlint` to CI or a pre-push hook so workflow input mismatches surface before push.

2. **The `flags:` block is now the single source of truth for any non-input gcloud flag.** Worth a comment at the top of the block (e.g. `# any gcloud run deploy flag goes here; top-level inputs are limited to the action's API`). Today it's implicit.

3. **Future asymmetry checks.** When adding a new `google-github-actions/*` step, the pattern from this fix should be: assume nothing about top-level inputs from sibling actions; consult the action's README explicitly. Worth adding to the runbook or a "patterns I used to misremember" section.

---

Issue #36, opened 2026-04-28
feat(ci): preview-deploy.yml uses NEON_PARENT_BRANCH for per-PR branch parent


va-worker:
## Problem Statement

`preview-deploy.yml` calls `neondatabase/create-branch-action` to make a `pr-{number}` branch for each PR's preview. Today it sets `parent: main` (line 58 in the current file). Under the branch-per-attendee model from #33, **a participant's PR should branch from THEIR branch, not from `main`** — otherwise their preview database doesn't see the schema changes they've already made on their working branch.

Concrete failure:

- Alice's branch (`alice`) has `pgvector` extension enabled and a `documents` table she added via Alembic.
- She opens a PR adding semantic search to her FastAPI app.
- `preview-deploy.yml` creates Neon branch `pr-42` from `main` (no pgvector, no `documents` table).
- Her PR's preview deploy fails at startup because the schema doesn't match what her code expects.

The fix: `preview-deploy.yml` reads `NEON_PARENT_BRANCH` from the fork's GitHub Actions secrets and passes it to `neondatabase/create-branch-action` as the `parent` parameter. Each fork sets its own value (e.g. `alice` for alice's fork). When unset, fall back to `main` (current behavior, useful for forks that aren't part of a workshop).

**Parent Epic:** vibeacademy/agile-flow-gcp#2 — Workshop-ready hardening of the GCP fork
**Effort Estimate:** XS
**Priority:** P1 (workshop participants only hit this when they open PRs; the first day-of deploy from main works without this fix)
**Depends on:** #33 (provisioner sets up the per-attendee branch); #35 (CSV column)
**Note:** the facilitator runbook update (separate ticket) needs to tell facilitators to set `NEON_PARENT_BRANCH=<handle>` as a GitHub Actions secret on each participant's fork. This ticket is the workflow-side change only.

---

## A. Environment Context

- File: `.github/workflows/preview-deploy.yml`
- Existing line that needs to change (around line 58, exact location subject to drift):
  ```yaml
  - name: Create Neon branch
    id: neon
    uses: neondatabase/create-branch-action@v5
    with:
      project_id: ${{ secrets.NEON_PROJECT_ID }}
      branch_name: pr-${{ env.PR_NUMBER }}
      parent: main                  # ← change this
      ...
  ```
- The action accepts `parent` as either a branch name string or a branch ID. We use a name. The participant's branch is named whatever `NEON_PARENT_BRANCH` says.

---

[...]

PR #38, opened 2026-04-28
feat(scripts): add optional neon_branch column to roster.csv + wrapper plumbing

### LINKED ISSUE #35: feat(scripts): add optional neon_branch column to roster.csv + wrapper plumbing
## Problem Statement

The `roster.csv` schema today is:

```csv
handle,github_user,email,cohort
```

The Neon-branch-per-attendee work in #33 needs a way for the wrapper to know each participant's branch name. We could derive it from `handle` (e.g. always use `handle` as the branch name), but that conflates two concerns:

1. The handle appears in the GCP project ID (`af-{handle}-{cohort}`) and is constrained by GCP project-ID rules (lowercase, no underscores, etc.).
2. The Neon branch name has different rules (Neon allows underscores, longer names, etc.) and may need to differ — for example, if a participant returns to a second cohort, the facilitator might want to keep the same branch name across cohorts to preserve their data, while the GCP project ID rotates with the cohort.

Adding an optional `neon_branch` column with default-to-handle gives flexibility without forcing facilitators to learn a new column for the simple case.

**Effort Estimate:** XS
**Priority:** P0 (depends on by #33; #33 reads the column to compute branch names)
**Depends on:** None — this ticket is the foundation; #33 builds on it.
**Blocks:** #33

---

## A. Environment Context

- Files to modify:
  - MODIFY `scripts/provision-workshop-roster.sh` — read 5th column with awk-style fallback to column 1 when empty; export `NEON_BRANCH_NAME`
  - MODIFY `scripts/roster.example.csv` — add a 5th column header and example values (one row using the default, one row showing an explicit override)
  - MODIFY `scripts/workshop-setup.sh` — pre-flight check needs to accept either 4-column or 5-column header
  - MODIFY `scripts/workshop-teardown.sh` — same header-flexibility (it only reads handle + cohort, but its CSV header validation must allow the new column)
  - MODIFY `docs/PLATFORM-GUIDE.md` — document the column under "Roster format"

---

## B. Guardrails

- The new column MUST be optional. Existing rosters with the 4-column header MUST continue to work — the wrapper falls back to using `handle` as `NEON_BRANCH_NAME`.
- Header validation accepts BOTH `handle,github_user,email,cohort` and `handle,github_user,email,cohort,neon_branch`. Anything else exits 2 as today.
- `NEON_BRANCH_NAME` MUST follow Neon's naming rules: 1-63 chars, alphanumeric + hyphens + underscores. Validate before exporting; if invalid, fail-fast on that row with a clear message.
- Do NOT change the project-ID derivation (`af-{handle}-{cohort}`). This ticket only adds plumbing for Neon, not GCP.
- Do NOT modify `provision-gcp-project.sh`. It receives `NEON_BRANCH_NAME` via env var; the inner script's logic is in #33.

---

## C. Happy Path

1. Facilitator's roster.csv is one of:

   ```csv
   handle,github_user,email,cohort
   alice,alice-gh,alice@example.com,2026-05
   ```

   (4 columns; `NEON_BRANCH_NAME=alice` derived from `handle`)

   OR

   ```csv
   handle,github_user,email,cohort,neon_branch
   alice,alice-gh,alice@example.com,2026-05,alice_personal
   ```

   (5 columns; `NEON_BRANCH_NAME=alice_personal` from explicit value)

2. Wrapper validates header against both shapes.
3. Per row, wrapper exports `NEON_BRANCH_NAME` to the inner script invocation.
4. Inner script (post-#33) reads `NEON_BRANCH_NAME` to create/look-up the Neon branch.

---

## D. Definition of Done

- [ ] `scripts/provision-workshop-roster.sh` accepts both 4-column and 5-column headers
- [ ] When `neon_branch` column is present and non-empty, that value is exported as `NEON_BRANCH_NAME`
- [ ] When the column is absent or empty for that row, `NEON_BRANCH_NAME` defaults to the row's `handle`
- [ ] Branch name is validated against Neon's pattern; invalid values exit 2 with a row-specific error message
- [ ] `scripts/workshop-setup.sh` pre-flight accepts both header shapes
- [ ] `scripts/workshop-teardown.sh` accepts both header shapes (it doesn't read the new column, just tolerates its presence)
- [ ] `scripts/roster.example.csv` shows both shapes (or one shape with the new column documented as optional)
- [ ] `docs/PLATFORM-GUIDE.md` documents the column
- [ ] `scripts/provision-workshop-roster.test.sh` adds a test for: 5-column input → exports the right value; 4-column input → defaults to handle; invalid neon_branch → exits 2
- [ ] All existing 14 wrapper-test assertions still pass

---

## Why P0

#33 cannot land without this. `provision-gcp-project.sh`'s new step reads `NEON_BRANCH_NAME`; the wrapper has to set it. Filing as a separate ticket so the diffs stay narrow and reviewable.

va-worker:
## Summary

Closes #35. **Foundation ticket — #33 builds on this.**

Adds an optional 5th column (`neon_branch`) to `roster.csv` and exports `NEON_BRANCH_NAME` per row to the inner provisioning script. The script-side Neon API integration in #33 will read it.

## CSV format

```csv
handle,github_user,email,cohort               (4 columns, unchanged behavior)
handle,github_user,email,cohort,neon_branch   (5 columns, with optional override)
```

- 5th column **empty or absent** → `NEON_BRANCH_NAME` defaults to `handle`
- 5th column **present and non-empty** → used as-is (after whitespace stripping)
- Invalid value (anything other than `^[A-Za-z0-9_-]{1,63}$`) → row fails fast with clear error

The 4-column shape continues to work — existing rosters need zero changes.

## Why an override exists

The default (`neon_branch = handle`) is correct for almost every workshop. The override matters when the same attendee returns for a second cohort: GCP project ID changes (`af-alice-2026-05` → `af-alice-2026-06`), but if they want the same Neon branch (continuity of data), the override gives it to them.

## Files changed

- `scripts/provision-workshop-roster.sh` — both header shapes accepted; loop reads 5 fields; default-to-handle; validate; forward `NEON_BRANCH_NAME`, `NEON_API_KEY`, `NEON_PROJECT_ID` to inner script
- `scripts/workshop-setup.sh` — pre-flight accepts both header shapes
- `scripts/workshop-teardown.sh` — header check accepts both; loop captures 5th field to prevent leak into `cohort`
- `scripts/roster.example.csv` — switched to 5-column shape with one default and one override
- `scripts/*.test.sh` (3 files) — 8 new assertions covering 5-column happy path, 4-column compatibility, invalid override, setup/teardown header acceptance
- `docs/PLATFORM-GUIDE.md` — Roster format section documents both shapes; corrects stale "github_user reserved for future use" note

## Test results

| Suite | Before | After |
|---|---|---|
| `provision-workshop-roster.test.sh` | 14/14 | 22/22 |
| `workshop-setup.test.sh` | 17/17 | 19/19 |
| `workshop-teardown.test.sh` | 14/14 | 16/16 |
| `provision-gcp-project.test.sh` (regression check) | 40/40 | 40/40 |
| **Total** | 85 | **97** |

12 new assertions; zero regressions.

## What this PR does NOT do

- No Neon API calls, no Secret Manager writes — that's #33
- No `NEON_PARENT_BRANCH` GitHub Actions secret integration — that's #36
- No facilitator runbook updates — that's `agile-flow-meta`#89

## Test plan

- [x] `bash -n` and `shellcheck` clean on all 6 modified scripts
- [x] `markdownlint` clean on `docs/PLATFORM-GUIDE.md`
- [x] All 4 test suites pass: 97/97 assertions
- [x] No regression on existing 85 assertions
- [ ] CI green (will verify after push)
- [ ] Real-GCP smoke at user's next nuke-and-fork

### DIFF
--- a/docs/PLATFORM-GUIDE.md
+++ b/docs/PLATFORM-GUIDE.md
@@ -327,7 +327,10 @@ discovering a missing auth token mid-loop.
 
 ### Roster format
 
-Create `roster.csv` with this exact header:
+`roster.csv` accepts two header shapes. Both have the same first four
+columns; the 5th is optional and used only by the Neon-branch automation.
+
+**Minimal (4 columns):**
 
 ```csv
 handle,github_user,email,cohort
@@ -335,16 +338,32 @@ alice,alice-gh,alice@example.com,2026-05
 bob,bob-gh,bob@example.com,2026-05
 ```
 
+**Extended (5 columns):**
+
+```csv
+handle,github_user,email,cohort,neon_branch
+alice,alice-gh,alice@example.com,2026-05,
+bob,bob-gh,bob@example.com,2026-05,bob_personal
+```
+
+Columns:
+
 - `handle` — short, lowercase, stable identifier; appears in the GCP project ID
-- `github_user` — reserved for future use (WIF binding, notification);
-  required in the row but not used by the current scripts
+- `github_user` — the participant's GitHub username; used to scope the
+  WIF binding to `<github_user>/agile-flow-gcp`
 - `email` — Google identity granted `roles/editor` on the new project
 - `cohort` — `YYYY-MM` of the workshop date; appears in the project ID
+- `neon_branch` *(optional, 5-column shape only)* — Neon branch name for
+  this attendee. When empty or absent, defaults to `handle`. Use the
+  override when the same person needs a stable branch across cohorts
+  (different `cohort`, same `neon_branch`). Must be 1-63 chars,
+  alphanumeric + hyphen + underscore.
 
 Project IDs follow the pattern `af-{handle}-{cohort}`. This shape is
 referenced from the facilitator runbook, the participant day-1 doc, and
 the dry-run checklist — do not change it. A working example lives at
-`scripts/roster.example.csv`.
+`scripts/roster.example.csv` (uses the 5-column shape with one default
+and one explicit override).
 
 ### Setup behavior
 
--- a/scripts/provision-workshop-roster.sh
+++ b/scripts/provision-workshop-roster.sh
@@ -15,11 +15,21 @@
 #   GCP_REGION           (default: us-central1) — passed through to inner script
 #   ARTIFACT_REPO        (default: agile-flow)  — passed through to inner script
 #   PROVISION_SCRIPT     (default: scripts/provision-gcp-project.sh) — for tests
+#   NEON_API_KEY         optional; forwarded to inner script for branch creation
+#   NEON_PROJECT_ID      optional; forwarded to inner script for branch creation
 #
-# CSV format (header required):
+# CSV format (header required, accepts either 4 or 5 columns):
 #   handle,github_user,email,cohort
 #   alice,alice-gh,alice@example.com,2026-05
 #   bob,bob-gh,bob@example.com,2026-05
+#
+#   handle,github_user,email,cohort,neon_branch        (5-column variant)
+#   alice,alice-gh,alice@example.com,2026-05,alice
+#   bob,bob-gh,bob@example.com,2026-05,bob_personal    (explicit branch override)
+#
+# When the optional `neon_branch` column is empty or absent, NEON_BRANCH_NAME
+# defaults to the row's `handle`. Use the override when the same person needs
+# a stable Neon branch across cohorts (different `cohort` value, same branch).
 #
 # Project IDs follow the pattern  af-{handle}-{cohort}  and are globally
 # unique. This is non-negotiable: the runbook, day-1 doc, and dry-run
@@ -68,12 +78,25 @@ if [[ ! -x "$PROVISION_SCRIPT" ]]; then
 fi
 
 # ── CSV header validation ────────────────────────────────────────────────
+#
+# The roster format accepts two header shapes:
+#   1. handle,github_user,email,cohort               (4 columns, original)
+#   2. handle,github_user,email,cohort,neon_branch   (5 columns, with Neon
+#                                                     branch override)
+#
+# When the 5th column is present and non-empty for a row, NEON_BRANCH_NAME
+# takes that value. Otherwise it defaults to the row's `handle` — which
+# matches the GCP project ID's handle component, so attendee branches are
+# named alice / bob / etc. by default.
 
-EXPECTED_HEADER="handle,github_user,email,cohort"
+EXPECTED_HEADER_4="handle,github_user,email,cohort"
+EXPECTED_HEADER_5="handle,github_user,email,cohort,neon_branch"
 ACTUAL_HEADER="$(head -n 1 "$ROSTER_CSV" | tr -d '\r')"
 
-if [[ "$ACTUAL_HEADER" != "$EXPECTED_HEADER" ]]; then
-  echo "ERROR: roster CSV header must be exactly: $EXPECTED_HEADER" >&2
+if [[ "$ACTUAL_HEADER" != "$EXPECTED_HEADER_4" && "$ACTUAL_HEADER" != "$EXPECTED_HEADER_5" ]]; then
+  echo "ERROR: roster CSV header must be one of:" >&2
+  echo "       $EXPECTED_HEADER_4" >&2
+  echo "       $EXPECTED_HEADER_5" >&2
   echo "       got: $ACTUAL_HEADER" >&2
   exit 2
 fi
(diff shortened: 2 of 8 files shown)


va-reviewer:
## PR Review — #38

### Requirements

- [x] Both 4-column and 5-column headers accepted across all three wrapper scripts (provision, setup, teardown)
- [x] `NEON_BRANCH_NAME` exported per row; defaults to `handle` when 5th column is absent or empty
- [x] Validation matches Neon's branch-name rules (`^[A-Za-z0-9_-]{1,63}$`); invalid values fail row-fast with a clear error
- [x] `NEON_API_KEY` and `NEON_PROJECT_ID` forwarded via env-var passthrough; never written to disk or logged
- [x] `roster-output.csv` schema unchanged — connection strings will go to Secret Manager (per #33), not flat file
- [x] No `provision-gcp-project.sh` change — Neon API integration is correctly deferred to #33
- [x] No regression — 85 existing assertions still pass + 12 new = 97/97

### Code Quality

- [x] **Defensive `${neon_branch:-}` expansion** is load-bearing because `set -u` is on and 4-column rows leave the variable unset by `read`. Correctly handled.
- [x] Comments explain the pattern in three places (header validation, loop, env-passthrough) without repetition
- [x] `read -r` field count behavior is correct: 5 named fields means a 6th column would silently bleed into field 5, but the header check rejects that shape upstream — defense-in-depth
- [x] `tr -d '[:space:]\r'` matches the existing strip pattern for `handle`, `cohort`, `email`
- [x] `shellcheck` clean across all 6 modified scripts

### Testing

- **Test coverage is strong:** 5-column happy path with explicit override + default-fallback; 4-column regression; invalid-input rejection; setup pre-flight; teardown header acceptance.
- **Stub-log assertion technique** (`grep -q "GCP_PROJECT_ID=...*NEON_BRANCH_NAME=..."`) cleanly proves env-var forwarding on the right invocation.
- **The error-message-update assertion change** (`"header must be exactly"` → `"header must be one of"`) catches the existing test that would've broken silently.
- All 4 test suites pass (provision-workshop-roster: 22/22; workshop-setup: 19/19; workshop-teardown: 16/16; provision-gcp-project regression: 40/40).
- All standard CI checks pass.

### Security

- [x] No injection vectors — validation regex blocks shell metacharacters; values pass via env-var assignment on the gcloud command line, not concatenated into shell commands
- [x] `NEON_API_KEY` is environment-forwarded; never persisted or logged
- [x] `roster-output.csv` does not gain a Neon-related column (no risk of accidental URI exposure in flat file)

### Recommendation

**GO**

Foundation ticket for the Neon-branch-per-attendee work. The diff is exactly the scope #35 promised — no more, no less. #33 can now consume `NEON_BRANCH_NAME` (and forwarded API credentials) without further wrapper changes. Ready to merge.

These are observations, not blockers. Worth filing as quick-fix follow-ups if any bite during the dry-run.

1. **Test 8 has a dead first invocation.** Lines write a "bad branch with spaces" roster, run the wrapper, then immediately overwrite the roster with `bad$value` and run again. The first run's result is discarded. The author kept it as a comment-by-example showing why whitespace alone isn't enough to invalidate. Cleaner: delete the first invocation, keep the inline comment. One-line cleanup.

2. **`tr -d '[:space:]'` on `neon_branch` strips internal spaces silently.** `"my branch"` becomes `"mybranch"` which passes validation. Surprising-but-not-wrong — the resulting Neon branch name is valid. A `[warn] neon_branch had whitespace stripped from "my branch" → "mybranch"` log line would surface the silent transform. Low priority.

3. **Validation regex `^[A-Za-z0-9_-]{1,63}$` is more permissive than Neon's documented `^[A-Za-z0-9][A-Za-z0-9_-]{0,62}$`** (must start with alphanumeric). A leading hyphen or underscore would pass the wrapper and fail mid-API-call in #33's code. Tightening to Neon's exact pattern would fail-fast earlier with a clearer error. Trivial change; not blocking because Neon's error is also clear.

4. **`roster.example.csv` switched to 5-column shape with one default and one override.** That's a deliberate breaking change for any tooling that parsed the old 4-column example. Practically: no such tooling exists. Worth keeping — the example should show the recommended shape going forward, and `PLATFORM-GUIDE.md` documents the minimal 4-column variant for users who want it.

5. **The "stripped value differs from raw" pattern bites in CSV imports from spreadsheets.** Excel/Sheets sometimes export trailing tabs or non-breaking spaces. The strip handles those, but the silent normalization could surprise. A future hardening would be to validate the *raw* value, then strip — that way `"alice "` (trailing space) errors clearly instead of silently becoming `"alice"`. Out of scope here.

---

PR #39, opened 2026-04-28
feat(scripts): provision Neon branch + database-url secret per attendee

### LINKED ISSUE #33: feat(scripts): provision Neon branch + database-url secret per attendee

va-worker:
## Summary

Closes #33. **Workshop-blocking P0; depends on #35 (already merged).**

Adds Step 5.7 to `provision-gcp-project.sh` that creates a Neon branch and writes its pooled connection URI to a `database-url` Secret Manager secret. Cloud Run mounts the secret as `DATABASE_URL` at runtime. After this lands, fresh forks no longer need a manual `echo url | gcloud secrets create` step — the wrapper takes 8 attendees from zero to 8 isolated Postgres branches in one command.

## What the new step does

```
input env: NEON_API_KEY, NEON_PROJECT_ID, NEON_BRANCH_NAME, GCP_PROJECT_ID, SA_EMAIL

  GET /projects/{NEON_PROJECT_ID}/branches    → find default branch ID
  POST /projects/.../branches                  → create attendee branch (or 409)
  GET /projects/.../connection_uri?pooled=true → fetch pooled URI
  gcloud secrets describe database-url         → check existence
  gcloud secrets [create | versions add]       → write URI
  gcloud secrets add-iam-policy-binding        → per-secret IAM
```

When any of the three Neon env vars is unset, the step skips and the footer prints the manual fallback. Real-GCP smoke deferred to your next nuke-and-fork (consistent with the rest of today's chain).

## Design decisions (confirmed in the ticket discussion)

1. **Schema baseline on `main` is empty.** Attendees run `alembic upgrade head` as their first command — that's a teaching moment, not a setup chore.
2. **Idempotent re-runs reuse existing branches.** No drift check against `main`. Workshops are short-lived enough that drift isn't real.
3. **Connection URIs are never logged.** Only written to Secret Manager. The script doesn't even echo them in error messages.

### Curl + Python for Neon API

The Neon API is JSON-only. `jq` isn't a system dependency I want to take, so the script uses `python3 -c "import json,sys; ..."` for parsing. Macros are inline because shellcheck struggles with complex heredocs; one-liner Python is fine. The pattern is:

```bash
parent_branch_id="$(echo "$branches_json" | python3 -c "
import json, sys
data = json.load(sys.stdin)
for b in data.get('branches', []):
    if b.get('default'):
        print(b['id'])
        sys.exit(0)
sys.exit('no default branch found')
" 2>&1)"
```

### Distinguishing 409 from real errors

`curl --fail-with-body` exits non-zero on 4xx/5xx. For the branch-create call we *want* to know about 409 (branch already exists, fall through to lookup) without halting. So that one call uses `--write-out '%{http_code}'` instead of `--fail`, and we branch on the captured code. The other two calls (list branches, get connection URI) use `--fail-with-body` because any non-2xx there is a real error.

### Pooled URI sanity check

The script logs `WARN: connection URI does not contain '-pooler.'` if the returned URI doesn't have the pooler host fragment. This is a soft check — the Neon API might one day change the host shape — but it surfaces the most likely silent failure (pattern #9: pooled vs direct).

### Footer rewrite

The `if [[ "$NEON_BRANCH_PROVISIONED" == "true" ]]` branch in the footer:
- Drops the manual `gcloud secrets create database-url ...` snippet
- Mentions `NEON_PARENT_BRANCH = $NEON_BRANCH_NAME` as the value to set on the participant's fork (forward-reference to #36)
- Keeps the existing manual flow as the `else` branch

## Tests

`scripts/provision-gcp-project.test.sh` gains tests 15-17:

- **Test 15** (NEON_API_KEY unset): asserts the skip log line and zero Neon API calls
- **Test 16** (branch absent): asserts both `[neon] branch created` AND `[create] database-url secret`, plus that `gcloud secrets create` was invoked AND IAM binding was granted
- **Test 17** (branch + secret already current): asserts the idempotent path — both skip lines logged AND `gcloud secrets create` was NOT called

All three tests use new stubs:

- `curl` stub matches by URL path and returns parametric responses (404, 201, 409, 500). The stub correctly handles `--output FILE` (writes body to file) and `--write-out '%{http_code}'` (writes code to stdout).
- `gcloud` stub extends the Step 5.5 stub with `secrets describe`, `secrets versions access latest`, `secrets create`, and `secrets add-iam-policy-binding`.

**Total: 106/106 assertions pass** across all 4 test files (49 inner + 22 wrapper + 19 setup + 16 teardown).

## Test runtime note

Tests 15-17 each take ~30s because the script's SA-propagation probe (added in #16) waits 5×6 seconds when its describe stub returns failure. The probe is correct for real-GCP use; for these tests the wait is wasted. Worth a small follow-up: support `WIF_PROBE_SLEEP` or similar env override in test mode. Not blocking — it's just slow, not broken.

## Test plan

- [x] 49/49 inner-script tests pass; 22/22 + 19/19 + 16/16 wrapper tests pass
- [x] markdownlint clean on PLATFORM-GUIDE.md
- [ ] CI green (will verify after push)
- [ ] Real-GCP smoke at user's next nuke-and-fork

### DIFF
--- a/docs/PLATFORM-GUIDE.md
+++ b/docs/PLATFORM-GUIDE.md
@@ -258,6 +258,40 @@ From the Neon Console, grab:
 
 ### Step 7: Create a Secret Manager Secret for the Database URL
 
+**As of `provision-gcp-project.sh` Step 5.7, this is automatic per
+project.** Set three env vars before running the provisioner:
+
+```bash
+NEON_API_KEY=neon_api_key_... \
+NEON_PROJECT_ID=dawn-mountain-12345 \
+NEON_BRANCH_NAME=alice \
+GCP_PROJECT_ID=af-alice-2026-05 \
+BILLING_ACCOUNT_ID=XXX-XXXX-XXXX \
+  ./scripts/provision-gcp-project.sh --create-project
+```
+
+Step 5.7 will:
+
+1. Create a Neon branch named `$NEON_BRANCH_NAME` (parented from the
+   project's `main` branch). If the branch already exists, reuse it.
+2. Fetch the branch's pooled connection URI via the Neon API.
+3. Create the `database-url` Secret Manager secret with that URI as
+   the value. If the secret already exists with a different value, add
+   a new version.
+4. Grant the deployer service account `roles/secretmanager.secretAccessor`
+   on the secret.
+
+The workshop wrapper (`provision-workshop-roster.sh`) reads
+`neon_branch` from each `roster.csv` row (defaulting to the row's
+`handle`) and exports `NEON_BRANCH_NAME` automatically. Facilitators
+running the canonical workshop flow don't need to manage this secret
+per project.
+
+#### Manual fallback (rarely needed)
+
+If you're running the inner script standalone or the env vars aren't
+set, the script's "Next steps" footer prints the manual command:
+
 ```bash
 echo -n "postgresql://user:pass@pooled-host/db" | \
   gcloud secrets create database-url \
@@ -265,8 +299,10 @@ echo -n "postgresql://user:pass@pooled-host/db" | \
     --project=YOUR_PROJECT_ID
 ```
 
-Use the **pooled** Neon connection string. Cloud Run exhausts direct
-connections fast because every revision instance opens its own pool.
+Use the **pooled** Neon connection string (the host should contain
+`-pooler.`). Cloud Run exhausts direct connections fast because every
+revision instance opens its own pool. See pattern #9 in
+`docs/PATTERN-LIBRARY.md`.
 
 ### Step 8: Configure GitHub Secrets and Variables
 
--- a/scripts/provision-gcp-project.sh
+++ b/scripts/provision-gcp-project.sh
@@ -15,6 +15,19 @@
 #   GCP_REGION           (default: us-central1)
 #   ARTIFACT_REPO        (default: agile-flow)
 #   BILLING_ACCOUNT_ID   (required if --create-project)
+#   GITHUB_USERNAME      (optional) Enables Step 5.5 (WIF setup). Pinned
+#                        to <user>/${WIF_REPO:-agile-flow-gcp}.
+#   WIF_REPO             (default: agile-flow-gcp) Override the repo name
+#                        WIF binds to. Workshop participants fork the
+#                        canonical template without renaming, so the
+#                        default fits.
+#   NEON_API_KEY         (optional) Enables Step 5.7 (Neon branch +
+#                        database-url Secret Manager). Required for
+#                        per-attendee branch automation.
+#   NEON_PROJECT_ID      (optional) Same: required for Step 5.7.
+#   NEON_BRANCH_NAME     (optional) Same: required for Step 5.7.
+#                        Wrapper sets this from the CSV's neon_branch
+#                        column (defaults to handle).
 #
 # Notes:
 # - This script is idempotent. Re-running skips resources that already exist.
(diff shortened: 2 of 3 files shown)


va-reviewer:
## PR Review — #39

### Requirements

- [x] Step 5.7 gated on `NEON_API_KEY`, `NEON_PROJECT_ID`, `NEON_BRANCH_NAME` all being set
- [x] Branch creation: POST first, fall through to lookup on 409 — race-safe (re-fetches if cached `branches_json` doesn't include the just-found branch)
- [x] Pooled URI fetched via `connection_uri?branch_id=...&pooled=true`; sanity-check warns when `-pooler.` is missing from the host
- [x] Secret Manager: create if absent; versions-add if different value; skip if same
- [x] Per-secret IAM binding granted to the deployer SA (defense-in-depth on top of Step 5's project-level binding)
- [x] Connection URIs **never logged** — only written to Secret Manager. `printf '%s'` (no trailing newline) avoids the URL-parsing bug from earlier in the session.
- [x] Footer rewritten to skip manual instructions when Step 5.7 ran; manual fallback retained for the env-unset path
- [x] Forward-references #36 (`NEON_PARENT_BRANCH`) so facilitators know what to set on the fork
- [x] No regression — 106/106 across all 4 test suites (49 inner + 22 + 19 + 16)

### Code Quality

- [x] Step 5.7 is ~180 lines including comments — linearly readable, no abstractions
- [x] Inline comments name the design decisions explicitly: idempotency policy, drift tolerance, race handling for the 409 path
- [x] `--fail-with-body` for GET (any 4xx/5xx is a real error); raw curl with `--write-out '%{http_code}'` for POST (so we can branch on 409)
- [x] `python3 -c` for JSON parsing avoids adding `jq` as a dependency
- [x] `set -e` is on; failures fail-fast cleanly
- [x] `shellcheck` clean

### Testing

- **Tests 15-17** cover all three branches via a `run_step5_7_test` helper that stubs both `curl` (Neon API) and `gcloud` (Secret Manager).
- **Curl stub correctly handles `--output FILE` + `--write-out '%{http_code}'`** — body to file, code to stdout. That's a non-trivial mock because the same call returns two pieces of information.
- **Test 16's strong assertions:** branch-created log, secret-create log, `gcloud secrets create` invoked, IAM binding granted. Catches three distinct regression risks at once.
- **Test 17 verifies idempotent skip:** secret already current → no `gcloud secrets create` call. The negative assertion is the load-bearing one — easy to regress to "always create new version" without it.
- All CI checks pass.

### Security

- [x] Connection URI never logged anywhere. Only written to Secret Manager via `printf '%s' | gcloud secrets create --data-file=-`. The `printf '%s'` (no `\n`) avoids the trailing-newline bug class.
- [x] `NEON_API_KEY` is in env, passed via `Authorization: Bearer`. Not logged.
- [x] Per-secret IAM binding scopes the deployer SA's access correctly.
- [x] No injection vectors — `gcloud` is argv-parsed; the curl URL is constructed from validated env vars (wrapper's regex from #35 catches shell metacharacters).
- [x] No `--no-verify`, no bypassed hooks.

### Recommendation

**GO**

P0 unblocking the production-deploy smoke. After merge: a fresh fork can deploy `main` end-to-end without manual `gcloud secrets create`. The implementation is careful (idempotency, race handling, security), the tests assert on the load-bearing invariants (no logging of secrets, idempotent skip path, IAM binding), and the design decisions (empty `main`, drift tolerance) are explicitly documented.

Ready to merge. The next ticket (#36, preview-deploy parent branch) reads `NEON_PARENT_BRANCH` which this PR's footer prints.

These are observations from a careful read; none justify a re-spin. Worth filing as follow-ups if any bite during dry-run.

1. **`python3` is now a hard runtime dependency** of `provision-gcp-project.sh` (4 call sites for JSON parsing). macOS ships it; some Linux distros don't. Worth adding a pre-flight check to `workshop-setup.sh`: `command -v python3` when the Neon env vars are set. Failure mode without it is a confusing `python3: command not found` mid-loop.

2. **JSON-parsing Python uses bash interpolation for `$NEON_BRANCH_NAME`** (lines 174 + 188 of the script). The wrapper (#35) validates the value against `^[A-Za-z0-9_-]{1,63}$` which excludes shell metacharacters and quotes, so today this is safe. But a direct caller bypassing the wrapper could pass a value with a single quote and get a Python syntax error instead of a clear validation message. Defensive fix: `NEON_BRANCH_NAME=... python3 -c "import os; target = os.environ['NEON_BRANCH_NAME']; ..."`. Not blocking.

3. **`--fail-with-body` requires curl ≥ 7.76 (2021).** macOS 13+ has curl 8; Ubuntu 22.04 has 7.81. Almost certainly fine. Worth a pre-flight check too.

4. **No `--max-time` on curl calls.** If Neon's API hangs, the script waits indefinitely. For an 8-attendee provisioning loop, one stuck row halts the whole loop. Adding `--max-time 30` to both curl helpers would bound the worst case. 30s is generous for Neon's API.

5. **`database_name=neondb` and `role_name=neondb_owner` are hardcoded** in the connection_uri request. These are Neon's defaults. If a facilitator named the database differently (`app_db`, etc.), the script wouldn't find a URI and the script would error. Optional `NEON_DATABASE_NAME` and `NEON_ROLE_NAME` env vars with current values as defaults would close this gap. Not urgent.

6. **The footer's "see #36" forward-reference will become stale once #36 ships.** Either drop the ticket reference now (since `NEON_PARENT_BRANCH` is the real env var) or track an "update this footer line" item in #36's DoD. Minor.

7. **Test 17 covers "exists, same value" but not "exists, different value."** Adding a Test 18 for the `versions add` path would round out coverage. The code path is symmetric so the regression risk is low, but explicit is better.

---

Issue #71, opened 2026-04-29
fix(provision): set PRODUCTION_DATABASE_URL GitHub secret during bootstrap


tck517:
## Problem Statement

`scripts/provision-gcp-project.sh` Step 7 (introduced in #57) auto-pushes 4 per-attendee GitHub Actions secrets to the participant's fork: `GCP_PROJECT_ID`, `GCP_SERVICE_ACCOUNT`, `GCP_WORKLOAD_IDENTITY_PROVIDER`, `NEON_PARENT_BRANCH`. It does **not** push `PRODUCTION_DATABASE_URL`, but `deploy.yml` reads that secret in two places:

- Line 93 — feeds the Alembic migration step before each production deploy
- Line 160 (post-#68) — supplies `DATABASE_URL` as a plain env var on the Cloud Run revision

When the bootstrap leaves the GitHub secret unset, both paths break:

- The migration step exits 0 with `PRODUCTION_DATABASE_URL not configured — skipping migrations.` Production Neon branch never gets the schema.
- The Cloud Run runtime starts with `DATABASE_URL=""` (empty). Since #68, this is no longer a Secret Manager mount — it's a plain env var passthrough that's now empty.

Symptom: the production app starts, hits 200 on `/api/health`, then 500s with `psycopg.errors.UndefinedTable: relation "todo" does not exist` on the first DB query. Looks like a schema bug; is actually a CI-secrets bug.

The downstream session at `tck517/tck517-app` hit this on first successful production load. Long-form analysis: `tck517/tck517-app:reports/session-journals/2026-04-29-upstream-handoff-bug4.md`.

**Parent Epic:** vibeacademy/agile-flow-gcp#48
**Effort Estimate:** S
**Priority:** P0 — workshop-blocking. Every fork's first production load fails until this is fixed. Compounded by #68 (which made the runtime path *also* depend on this secret being set).
**Surfaced:** 2026-04-29 dry-run on tck517/tck517-app, second handoff doc.

---

## A. Environment Context

- File: `scripts/provision-gcp-project.sh`
- Step 5.7 already mints the Neon pooled URI as `pooled_uri` and writes it to the Secret Manager secret `database-url`. Same value should land in the GitHub secret.
- Step 7 (the existing `gh secret set` block) is the right home — already gated on `gh` being on PATH and `GITHUB_REPOSITORY` set.
[...]

PR #73, opened 2026-04-30
feat(app): fail fast on missing/sqlite DATABASE_URL in production

### LINKED ISSUE #63: feat(app): fail fast on missing DATABASE_URL in production
## Problem Statement

`app/config.py` defaults `database_url: str = "sqlite:///./dev.db"`. In production, when the secret-mounted `DATABASE_URL` env var is missing, empty, or unreadable, the runtime silently constructs a SQLite engine. The first DB-touching request then 500s with `sqlite3.OperationalError: no such table: todo` — a downstream symptom that hides the real config bug.

The 2026-04-29 dry-run on tck517/tck517-app hit this: revisions 00003–00004 ran the FastAPI image but logged the SQLite error because the secret mount wasn't taking effect. By the time it shows up in error reports it looks like a schema bug, not a config bug.

The validator added in PR #55 (psycopg3 normalization) doesn't cover this case — it only rewrites `postgresql://` to `postgresql+psycopg://`. An empty/missing/sqlite-shaped URL isn't normalized, just used as-is.

**Effort Estimate:** S
**Priority:** P1 — defense-in-depth. The framework-level fixes in #61 and #62 should make this unreachable, but the seatbelt is cheap and would have made the dry-run incident a 30-second diagnosis.
**Surfaced:** 2026-04-29 dry-run on tck517/tck517-app.

---

## A. Environment Context

- File: `app/config.py`
- Existing tests: `tests/test_config.py` (created in PR #55)
- Related env var: `ENVIRONMENT` (defaults to `"development"`; `deploy.yml` sets it to `"production"`)

---

## B. Guardrails

- Fail at startup, not on first DB request. The earlier the failure, the less ambiguous the diagnosis.
- Only fail when `ENVIRONMENT == "production"` AND `database_url` is empty OR sqlite-shaped. Local dev (default `ENVIRONMENT=development`) MUST continue to work with the SQLite fallback.
- Error message must name BOTH the symptom (DATABASE_URL is empty/sqlite) and the likely cause (Secret Manager mount missing or wrongly named).
- Don't validate at import time — keep it in the `Settings` validator/model so test injection still works.

---

## C. Happy Path

In `app/config.py`, add a Pydantic model validator that runs after field validation:

```python
from pydantic import model_validator
# ... existing imports ...

class Settings(BaseSettings):
    # ... existing fields and validators ...

    @model_validator(mode="after")
    def _refuse_sqlite_in_production(self):
        if self.environment == "production":
            if not self.database_url:
                raise ValueError(
                    "DATABASE_URL is empty in production. "
                    "Check that the Secret Manager secret 'database-url' is "
                    "mounted in deploy.yml and the Cloud Run service has "
                    "secretAccessor on it."
                )
            if self.database_url.startswith("sqlite"):
                raise ValueError(
                    f"DATABASE_URL is SQLite in production: {self.database_url!r}. "
                    "Production must use Postgres. Likely cause: the secret-mounted "
                    "DATABASE_URL env var didn't override the dev default."
                )
        return self
```

After this lands, a misconfigured production deploy fails fast at startup with a clear message — Sentry sees a `ValueError` from config, not a `sqlite3.OperationalError` from a query handler.

---

## D. Definition of Done

- [ ] `app/config.py` has the `_refuse_sqlite_in_production` model validator
- [ ] `tests/test_config.py` adds 3+ cases:
  - production + empty DATABASE_URL → ValueError
  - production + sqlite:/// → ValueError
  - development + empty DATABASE_URL → no error (SQLite default applies)
- [ ] `uv run pytest` clean
- [ ] `uv run mypy app/` clean
- [ ] Manual verification: deploy with the secret unmounted, observe a clear `ValueError` in Cloud Run logs at startup (not a `sqlite3.OperationalError` on first request)

---

## Why P1

Defense-in-depth. The framework fixes (#61 + #62) should make this unreachable in practice, but config-defaults that silently mask production bugs are exactly what the pattern-library calls out as silent-skip cascade roots. Cheap, low-risk, high-MTTR-reduction.

va-worker:
## Summary

Closes #63. Adds a Pydantic `model_validator` to `Settings` that raises `ValueError` at construction time when `ENVIRONMENT=production` and `DATABASE_URL` is empty or sqlite-shaped.

## Why

The dev default `sqlite:///./dev.db` silently masks production config bugs. When a Secret Manager mount fails to override it, the runtime constructs a SQLite engine and the first DB-touching request 500s with `sqlite3.OperationalError: no such table: todo` — looks like a schema bug, is actually a config bug.

The 2026-04-29 dry-run on `tck517/tck517-app` burned ~30 min chasing this before finding the unmounted secret. Failing fast at startup with a named cause turns this into a 30-second diagnosis: a `ValueError` from config beats a `sqlite3.OperationalError` from a query handler every time.

## How this fits with #71

| Layer | Goal | Mechanism |
|---|---|---|
| #71 (already merged) | Make the bug unreachable | Auto-push `PRODUCTION_DATABASE_URL` so the secret mount has data |
| #63 (this PR) | Catch the bug if #71's preconditions are bypassed | Refuse to construct `Settings` with empty/sqlite URL in prod |

#63 is defense-in-depth — this validator should be unreachable in practice. It catches manual provisioning, drifted forks, secret rotation gone wrong, and any future regression in #71's auto-push path.

## What changed

| File | Change |
|---|---|
| `app/config.py` | New `_refuse_sqlite_in_production` `@model_validator(mode=\"after\")`. Gated on `environment == \"production\"`; non-production paths return `self` unchanged. Raises with a named cause ("Secret Manager secret 'database-url' is mounted in deploy.yml" / "didn't override the dev default") so the Cloud Run log entry points at the actual fix location. |
| `tests/test_config.py` | 4 new cases. Uses `monkeypatch` to control env vars per-test (no leakage between cases). |

## Tests

| Case | Expected |
|---|---|
| production + empty `DATABASE_URL` | `ValueError("DATABASE_URL is empty in production…")` |
| production + `sqlite:///./dev.db` | `ValueError("DATABASE_URL is SQLite in production…")` |
| development + empty `DATABASE_URL` | No raise — SQLite default applies |
| production + valid Postgres URL | No raise — normalized to `postgresql+psycopg://…` and accepted |

Existing 5 normalization tests still pass.

```
ruff   ✓
mypy   ✓ (10 source files)
pytest ✓ 17/17 (was 13)
```

## Test plan

- [ ] CI green
- [ ] After merge: deploy `main` to a fork with the secret intentionally unmounted. Confirm the Cloud Run revision exits at startup with the new `ValueError` in logs (not a `sqlite3.OperationalError` on first request).

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -6,8 +6,9 @@ or Cloud Run Secret Manager mounts.
 """
 
 from functools import lru_cache
+from typing import Self
 
-from pydantic import field_validator
+from pydantic import field_validator, model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 
@@ -32,6 +33,30 @@ class Settings(BaseSettings):
         if v.startswith("postgresql://"):
             return v.replace("postgresql://", "postgresql+psycopg://", 1)
         return v
+
+    @model_validator(mode="after")
+    def _refuse_sqlite_in_production(self) -> Self:
+        # Defense-in-depth: the dev-default sqlite URL must never reach
+        # a production runtime. If the secret-mounted DATABASE_URL is
+        # missing or the mount didn't override the default, fail at
+        # startup with a clear message rather than 500ing on the first
+        # DB query with `no such table: todo`. See #63 / #71.
+        if self.environment != "production":
+            return self
+        if not self.database_url:
+            raise ValueError(
+                "DATABASE_URL is empty in production. "
+                "Check that the Secret Manager secret 'database-url' is "
+                "mounted in deploy.yml and the Cloud Run service has "
+                "secretAccessor on it."
+            )
+        if self.database_url.startswith("sqlite"):
+            raise ValueError(
+                f"DATABASE_URL is SQLite in production: {self.database_url!r}. "
+                "Production must use Postgres. Likely cause: the secret-mounted "
+                "DATABASE_URL env var didn't override the dev default."
+            )
+        return self
 
 
 @lru_cache
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,9 +1,14 @@
 """Tests for app.config.Settings.
 
-Pins the contract that database_url is normalized to use the psycopg3
-driver scheme. SQLAlchemy resolves the bare `postgresql://` scheme to
-psycopg2 (not installed in this project), so without normalization
-every Neon-backed deploy would ImportError on first DB use.
+Pins two contracts:
+  1. database_url is normalized to use the psycopg3 driver scheme.
+     SQLAlchemy resolves the bare `postgresql://` scheme to psycopg2
+     (not installed in this project), so without normalization every
+     Neon-backed deploy would ImportError on first DB use.
+  2. In production, an empty or sqlite-shaped DATABASE_URL fails fast
+     at Settings construction (model validator) — turns the silent
+     "first request 500s with no such table: todo" failure mode into
+     a loud ValueError at startup. See #63.
 """
 
 import pytest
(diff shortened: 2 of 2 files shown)


va-reviewer:
## PR Review — #73

### Requirements
- [x] Acceptance criteria from #63 are met (validator added; 4 test cases — DoD asked for 3+; ruff + mypy + pytest clean)
- [x] Feature works end-to-end — independently verified that pydantic-settings v2.12 actually populates `database_url=""` from an empty env var (pydantic does NOT silently fall back to the default for empty strings), so the validator fires on the realistic failure mode
- [x] No scope creep — single `model_validator(mode="after")` plus tests; no touch on `app/main.py`, no behavior change for non-production paths

### Code Quality
- [x] Validator placement is correct: `model_validator(mode="after")` runs after `field_validator("database_url")`. Order: env-var read → field validator (psycopg3 rewrite, no-op for empty/sqlite) → model validator (catches empty/sqlite). Tested empirically against pydantic 2.12.5 / pydantic-core 2.41.5
- [x] Returns `Self` correctly typed for mypy; uses `typing.Self` (Python 3.11+; project pins 3.12 so this is fine)
- [x] Non-production short-circuit (`if self.environment != "production": return self`) is the cleanest possible gate — local dev with the default `sqlite:///./dev.db` still constructs fine
- [x] Error messages name BOTH the symptom (`"empty in production"` / `"is SQLite in production"`) AND the likely cause (`"Secret Manager secret 'database-url' is mounted in deploy.yml"` / `"didn't override the dev default"`). Matches B. Guardrails of the ticket exactly
- [x] Inline comment explains the WHY (the silent failure mode the validator prevents) and cross-references both #63 and #71 — future readers see the relationship between the two layers
- [x] No over-engineering: no separate validator class, no settings subclassing, no env-specific imports — just a 19-line method on the existing Settings model

### Testing
- [x] 4 cases cover the contract: prod+empty, prod+sqlite, dev+empty (the negative case — must NOT raise), prod+postgres (happy path also verifies field-validator chaining is preserved)
- [x] Uses `monkeypatch.setenv` / `monkeypatch.delenv` — no env-var leakage between tests, no manual cleanup
- [x] Each test imports `Settings` fresh (`from app.config import Settings` inside the test body) so module-level `lru_cache` on `get_settings` doesn't matter — and tests instantiate `Settings()` directly anyway, which bypasses the cache
- [x] `match=` patterns on the ValueError assertions verify the user-visible message, not just the exception class — meaningful regression guard for the helpful error text
- [x] Existing 5 normalization tests still pass; full suite 17/17 green
- [x] All CI checks green (lint, typecheck, build, test, python, node, lint-agent-policies)

### Security
- [x] No hardcoded secrets — test fixtures use `u:p@h.neon.tech` which is non-sensitive placeholder syntax
- [x] No injection vectors — strings are compared with `startswith` and emptiness, never expanded
- [x] Sensitive data handling: the `ValueError` for the sqlite case includes the URL via `f"...{self.database_url!r}..."`. For the only realistic prod failure (the dev-default `sqlite:///./dev.db`) this is fine — no credentials. A theoretical edge case where someone sets `DATABASE_URL=sqlite:///some-secret-string` in prod would log that string to Cloud Run, but (a) sqlite URLs don't carry credentials, (b) any such value is by definition not a real secret, (c) the error needs to be actionable. Not blocking
- [x] The empty-DATABASE_URL case does NOT echo the value (it's empty)
- [x] Dependencies unchanged — `model_validator` is part of the already-pinned `pydantic>=2.x`

### Recommendation
**GO**

Textbook defense-in-depth. Adds the loudest-possible failure mode for the silent-mask bug that #71 just made unreachable, with sharp error messages that point at the actual fix location. Tests cover the contract precisely, including the non-trivial dev-still-works negative case. Integration with the existing field validator is correct (verified empirically against pydantic 2.12). 17/17 green; CI green.

None.

- The `from app.config import Settings` inside each test body is necessary to pick up monkeypatched env (correct pattern), but the existing `_make_settings` helper at the top of the file does the same. Future refactor opportunity: a parametrized fixture that takes `(env, db_url)` and yields a constructed `Settings` (or the raised exception) would dedupe the 4 new tests into one parametrize block. Pure cleanup; out of scope.
- The error message for the empty case names `deploy.yml` and `secretAccessor`. If the eventual platform layer ever changes (e.g., a non-Cloud-Run target), this message becomes stale. Not actionable now, but worth a `# update if deploy target changes` note in a future cleanup pass.
- Minor: `self.database_url.startswith("sqlite")` would also catch `sqlitefoo://` (vanishingly unlikely real URL). Tightening to `startswith("sqlite:")` would be more precise. Pure pedantry; current code is correct for every realistic input.

PR #80, opened 2026-04-30
fix(scripts): activate pre-push hook in provisioner; doctor fail on dormant hook

### LINKED ISSUE #77: fix(scripts): provisioner activates pre-push hook so fresh forks don't ship with quality gate dormant
## Problem Statement

The pre-push hook at `scripts/hooks/pre-push` exists and is correct, but **fresh forks ship with it inactive.** Git's `core.hooksPath` config is per-clone, not per-repo. Nothing in the bootstrap path runs the activation command:

```bash
git config --local core.hooksPath scripts/hooks
```

Result: every fork can `git push origin main` without lint or tests running. The CLAUDE.md rule "All tests must pass before pushing. Never use `--no-verify`" is unenforceable when there is no hook to skip.

Surfaced 2026-04-30 dry-run on `tck517/tck517-app` while pushing the Issue #2 fix — `git push origin main` succeeded with no "Running pre-push checks…" output. Confirmed `git config --get core.hooksPath` was empty. Long-form analysis in `tck517/tck517-app:docs/UPSTREAM-HANDOFF.md` Issue #3.

`scripts/doctor.sh:244-249` would warn about this, but doctor is opt-in and a fresh fork has no signal to run it.

**Effort Estimate:** XS
**Priority:** P1 — silently weakens the quality gate on every fork. Doesn't directly cause production breakage but removes the safety net that catches the next class of bug.

---

## A. Environment Context

- File: `scripts/provision-gcp-project.sh` — fork-bootstrap entry point. Add the activation early (Step 0 before any GCP calls so even a partial provisioner run leaves the hook active).
- File: `scripts/doctor.sh:244-249` — currently `warn`s on missing hooksPath; upgrade to `fail` so `/doctor` is a hard error.
- File: `docs/GETTING-STARTED.md` — does not mention the hook today; add a "first commands after clone" section.
- The hook itself (`scripts/hooks/pre-push`) is good and does not need changes — well-written, language-detecting, defense-in-depth-aware. The problem is purely activation.

---

## B. Guardrails

- Use `--local` so a user's global `core.hooksPath` (e.g. for husky) is overridden only inside this repo.
- Activation must run BEFORE any failable GCP call so a partial run still leaves the hook active. Step 0 / pre-flight is the right home.
- Do not require `gh` or any other tool for activation — `git config` is part of every git install.
- The doctor.sh `fail`-not-`warn` upgrade should only fail when `scripts/hooks/pre-push` exists (i.e. the fork is using this template's hook). Forks that legitimately disabled the hook for their own reasons shouldn't be blocked.
- Don't auto-`chmod +x` the hook — `git update-index --chmod=+x` was used when it was added; clones inherit the bit. If a clone strips it, that's a separate platform-specific bug (Windows) outside this ticket's scope.

---

## C. Happy Path

In `scripts/provision-gcp-project.sh`, add after the existing `set -uo pipefail` and before any GCP call:

```bash
# ── Step 0: Activate the in-repo pre-push hook ─────────────────────────
#
# core.hooksPath is per-clone, not versioned with the repo. Without
# this, the working pre-push hook is dormant and a fresh fork ships
# with its quality gate inactive. See #<this-issue>.

if [[ -f scripts/hooks/pre-push ]] && [[ -d .git ]]; then
  current="$(git config --local --get core.hooksPath 2>/dev/null || true)"
  if [[ "$current" != "scripts/hooks" ]]; then
    git config --local core.hooksPath scripts/hooks
    echo "[hook] Activated pre-push hook (lint + tests will run before every push)"
  fi
fi
```

In `docs/GETTING-STARTED.md`, add a "Step 0: Activate quality gate" before the first GCP step:

```markdown
After cloning, activate the pre-push hook:

```bash
git config --local core.hooksPath scripts/hooks
```

This runs lint and tests automatically before every push. Without it, broken
code can reach `main` and trigger a failing production deploy.
```

In `scripts/doctor.sh:244-249`, replace the `warn` with a `fail`:

```diff
-    warn "Git Config" "core.hooksPath is '${hooks_path:-unset}' (expected scripts/hooks)" "git config --local core.hooksPath scripts/hooks"
+    fail "Git Config" "core.hooksPath is '${hooks_path:-unset}' (expected scripts/hooks); pre-push hook is dormant" "git config --local core.hooksPath scripts/hooks"
```

After this lands, a fresh fork running `provision-gcp-project.sh` has its hook active automatically. Forks that bypass the provisioner get the activation step from GETTING-STARTED. `/doctor` fails loud if the hook is dormant.

---

## D. Definition of Done

- [ ] `scripts/provision-gcp-project.sh` activates `core.hooksPath` early (before any GCP call), idempotent (no-op when already set), only when `scripts/hooks/pre-push` exists
- [ ] `scripts/doctor.sh` reports `fail` (not `warn`) when `core.hooksPath` is unset and `scripts/hooks/pre-push` exists
- [ ] `docs/GETTING-STARTED.md` opens with the `git config` command
- [ ] `provision-gcp-project.test.sh` extended: assert that running the script on a clean checkout sets `core.hooksPath` to `scripts/hooks`
- [ ] All 6 test suites green
- [ ] shellcheck clean

---

## Why P1

The pre-push hook would have caught Python-level breakage (import errors, type errors, broken tests, ruff violations). Any of those shipping to `main` triggers a production deploy. The fork has been operating without its quality gate the entire time, and nobody noticed because the hook silently skips rather than failing visibly. Fix one config call; remove an entire class of "broke main, didn't see it locally" failures.

va-worker:
## Summary

Closes #77. The pre-push hook at `scripts/hooks/pre-push` was correct, but `core.hooksPath` is per-clone (not versioned with the repo) and nothing in the bootstrap path activated it. Result: every fresh fork shipped with its quality gate dormant — `git push origin main` skipped lint and tests silently.

## Why

Surfaced 2026-04-30 dry-run on `tck517/tck517-app` while pushing the #76 fix. `git push` succeeded with no "Running pre-push checks…" output; `git config --get core.hooksPath` confirmed empty. The CLAUDE.md rule "Never use `git push --no-verify`" was structurally unenforceable when there was no hook to skip.

The hook itself doesn't need changes — well-written, language-detecting, defense-in-depth-aware. Pure activation gap.

## What changed

| File | Change |
|---|---|
| `scripts/provision-gcp-project.sh` | New **Step 0** between the banner and Step 1, BEFORE any failable GCP call. Runs `git config --local core.hooksPath scripts/hooks` when `.git/` and `scripts/hooks/pre-push` both exist. Idempotent (skips when already set). Position guarantees that even a partial provisioner run (failing later on missing GCP creds) still leaves the quality gate active. |
| `scripts/doctor.sh` lines 244-253 | Split the existing single `warn` into two branches: **FAIL** when the hook file is present but `core.hooksPath` is unset (dormant gate, the case #77 cares about); **WARN** when the hook file is also absent (fork that legitimately disabled the hook). Pre-existing `pass`-when-set branch unchanged. |
| `docs/GETTING-STARTED.md` Step 1 | Adds the `git config` command immediately after `cd my-app` in the clone block, with a one-paragraph explanation. Covers users who clone but bypass the provisioner. |

## Tests

New **Test 28** in `scripts/provision-gcp-project.test.sh` (4 assertions):

- `core.hooksPath` is `scripts/hooks` after the script runs in a fresh git repo
- Activation message `[hook] Activated pre-push hook` is logged on first run
- Idempotent: no activation message on re-run when already set
- Silent skip when `scripts/hooks/pre-push` is absent (no spurious log)

The test creates a temp git repo, places a placeholder `scripts/hooks/pre-push`, runs the provisioner with a stub `gcloud` (Step 0 fires before any GCP call so the stub doesn't matter for the assertion).

| Suite | Before | After |
|---|---|---|
| `provision-gcp-project.test.sh` | 93 | **97** |
| `provision-workshop-roster.test.sh` | 32 | 32 |
| `workshop-setup.test.sh` | 21 | 21 |
| `workshop-teardown.test.sh` | 18 | 18 |
| `ensure-github-account.test.sh` | 8 | 8 |
| `diagnose-cloudrun.test.sh` | 35 | 35 |
| **Total** | **207** | **211** |

`pytest` 17/17 green. shellcheck clean on the three changed shell files. (One pre-existing SC2034 warning in `doctor.sh:199` is unrelated to this PR; CI excludes SC2034 via `validate-scripts.sh`.)

## Verification post-merge

- [ ] CI green on this PR
- [ ] After merge: re-run the 2026-04-30 dry-run scenario. `git config --local --get core.hooksPath` returns `scripts/hooks` after `provision-gcp-project.sh` runs.
- [ ] Make a trivial change that breaks ruff. `git push` fails with "Pre-push checks failed".
- [ ] `scripts/doctor.sh` reports the hook check as `[PASS]`. (When deliberately unset for testing: reports `[FAIL]` with the actionable command.)

### DIFF
--- a/docs/GETTING-STARTED.md
+++ b/docs/GETTING-STARTED.md
@@ -57,8 +57,14 @@ Now clone it locally:
 cd ~/projects   # or wherever you keep repos
 git clone https://github.com/your-name/my-app.git
 cd my-app
+git config --local core.hooksPath scripts/hooks
 npm install
 ```
+
+The `git config` line activates the pre-push quality gate. Without it,
+broken code can reach `main` and trigger a failing production deploy.
+The setting is per-clone, not stored in the repo, so every fresh clone
+needs this command once.
 
 **You should see:** Dependencies installed with no errors.
 
--- a/scripts/doctor.sh
+++ b/scripts/doctor.sh
@@ -241,12 +241,16 @@ else
     fail "Git Config" "user.email not set" "git config --global user.email \"you@example.com\""
 fi
 
-# core.hooksPath (WARN)
+# core.hooksPath (FAIL when the hook file exists — pre-push gate dormant)
 hooks_path=$(git config --local core.hooksPath 2>/dev/null || true)
 if [ "$hooks_path" = "scripts/hooks" ]; then
     pass "Git Config" "core.hooksPath set to scripts/hooks"
+elif [ -f "scripts/hooks/pre-push" ]; then
+    # Hook exists but is dormant — quality gate is not running. See #77.
+    fail "Git Config" "core.hooksPath is '${hooks_path:-unset}'; pre-push hook is dormant" "git config --local core.hooksPath scripts/hooks"
 else
-    warn "Git Config" "core.hooksPath is '${hooks_path:-unset}' (expected scripts/hooks)" "git config --local core.hooksPath scripts/hooks"
+    # No hook file in repo — nothing to activate. Just note it.
+    warn "Git Config" "core.hooksPath is '${hooks_path:-unset}' (no scripts/hooks/pre-push present)" "git config --local core.hooksPath scripts/hooks"
 fi
 
 # pre-push exists + executable (WARN)
--- a/scripts/provision-gcp-project.sh
+++ b/scripts/provision-gcp-project.sh
@@ -141,6 +141,30 @@ retry_eventual_consistency() {
   return 1
 }
 
+# ── Step 0: Activate the in-repo pre-push hook ──────────────────────────
+#
+# `core.hooksPath` is per-clone, not versioned with the repo. Without
+# this, the working pre-push hook at `scripts/hooks/pre-push` is dormant
+# on every fresh clone — `git push` skips lint and tests silently. The
+# CLAUDE.md rule "Never use `git push --no-verify`" is unenforceable
+# when there is no hook to skip.
+#
+# Run this BEFORE arg-parsing, input validation, and any GCP call so
+# that a partial run — including the most common day-1 failure path,
+# `provision-gcp-project.sh` invoked without GCP_PROJECT_ID set — still
+# leaves the quality gate active. Idempotent: no-op when already set.
+# Uses --local so a user's global hooksPath (e.g. for husky) is
+# overridden only inside this repo. See #77.
+
+if [[ -d .git ]] && [[ -f scripts/hooks/pre-push ]]; then
+  current_hooks_path="$(git config --local --get core.hooksPath 2>/dev/null || true)"
+  if [[ "$current_hooks_path" != "scripts/hooks" ]]; then
+    git config --local core.hooksPath scripts/hooks
+    echo "[hook] Activated pre-push hook (lint + tests will run before every push)"
+    echo ""
+  fi
+fi
+
 CREATE_PROJECT=false
 WITH_SA_KEY=false
 
@@ -171,28 +195,6 @@ echo "  Repo:     $ARTIFACT_REPO"
 echo "  Create:   $CREATE_PROJECT"
 echo "  SA key:   $WITH_SA_KEY"
 echo ""
-
-# ── Step 0: Activate the in-repo pre-push hook ──────────────────────────
-#
-# `core.hooksPath` is per-clone, not versioned with the repo. Without
-# this, the working pre-push hook at `scripts/hooks/pre-push` is dormant
-# on every fresh clone — `git push` skips lint and tests silently. The
-# CLAUDE.md rule "Never use `git push --no-verify`" is unenforceable
-# when there is no hook to skip.
-#
-# Run this BEFORE any failable GCP call so even a partial provisioner
-# run leaves the quality gate active. Idempotent: no-op when already
-# set. Uses --local so a user's global hooksPath (e.g. for husky) is
-# overridden only inside this repo. See #77.
-
-if [[ -d .git ]] && [[ -f scripts/hooks/pre-push ]]; then
-  current_hooks_path="$(git config --local --get core.hooksPath 2>/dev/null || true)"
-  if [[ "$current_hooks_path" != "scripts/hooks" ]]; then
-    git config --local core.hooksPath scripts/hooks
-    echo "[hook] Activated pre-push hook (lint + tests will run before every push)"
-    echo ""
-  fi
-fi
 
 # ── Step 1: Create project (optional) ────────────────────────────────────
 
--- a/scripts/provision-gcp-project.sh
+++ b/scripts/provision-gcp-project.sh
@@ -172,6 +172,28 @@ echo "  Create:   $CREATE_PROJECT"
 echo "  SA key:   $WITH_SA_KEY"
 echo ""
 
+# ── Step 0: Activate the in-repo pre-push hook ──────────────────────────
+#
+# `core.hooksPath` is per-clone, not versioned with the repo. Without
+# this, the working pre-push hook at `scripts/hooks/pre-push` is dormant
+# on every fresh clone — `git push` skips lint and tests silently. The
+# CLAUDE.md rule "Never use `git push --no-verify`" is unenforceable
+# when there is no hook to skip.
+#
+# Run this BEFORE any failable GCP call so even a partial provisioner
+# run leaves the quality gate active. Idempotent: no-op when already
+# set. Uses --local so a user's global hooksPath (e.g. for husky) is
+# overridden only inside this repo. See #77.
+
+if [[ -d .git ]] && [[ -f scripts/hooks/pre-push ]]; then
+  current_hooks_path="$(git config --local --get core.hooksPath 2>/dev/null || true)"
+  if [[ "$current_hooks_path" != "scripts/hooks" ]]; then
+    git config --local core.hooksPath scripts/hooks
+    echo "[hook] Activated pre-push hook (lint + tests will run before every push)"
+    echo ""
+  fi
+fi
+
 # ── Step 1: Create project (optional) ────────────────────────────────────
 
 if [[ "$CREATE_PROJECT" == "true" ]]; then
--- a/scripts/provision-gcp-project.test.sh
+++ b/scripts/provision-gcp-project.test.sh
@@ -2127,6 +2127,97 @@ else
   FAIL=$((FAIL + 1))
 fi
 
+# ── Test 28: Step 0 activates core.hooksPath when pre-push hook exists ──
+#
+# Per #77: every fresh fork must leave the pre-push hook active after the
+# provisioner runs. core.hooksPath is per-clone and not auto-set by git
+# clone, so without this step the working hook is dormant. Run the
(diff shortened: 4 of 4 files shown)


va-reviewer:
## PR Review — #80

### Requirements
- [x] Provisioner activates `core.hooksPath` (idempotent, gated on `.git/` and hook file existing)
- [x] doctor.sh splits warn into FAIL (hook present but dormant) and WARN (no hook file) — three-branch logic is correct
- [x] GETTING-STARTED documents the `git config` command (placement is inside Step 1's clone block, which is the first concrete command — minor scope-of-"opens" call but acceptable)
- [x] Test 28 added in provision-gcp-project.test.sh with 4 assertions
- [ ] **Activation does NOT run "before any failable GCP call" as the inline comment claims.** See Required Changes below.

### Code Quality
- [x] Step 0 block is well-commented; cross-references #77; explains the per-clone semantics, the `--local` choice, and the idempotency
- [x] `[[ -d .git ]] && [[ -f scripts/hooks/pre-push ]]` gate is correct — no-ops in non-fork contexts (e.g. running the script from a `tar` extract without git history) and when the hook file is absent
- [x] Inline doctor.sh comments in the new branches reference #77, future-readable
- [x] GETTING-STARTED prose accurately explains why per-clone matters
- [ ] **Comment at the top of Step 0 says "BEFORE any failable GCP call so even a partial provisioner run leaves the quality gate active" but the actual placement is AFTER the `GCP_PROJECT_ID` required-check (line 159) and AFTER the argument-parsing `exit 1` paths.** A user who runs the script without `GCP_PROJECT_ID` set — the most common "first run, didn't read the docs" path — hits `exit 1` at line 162 and Step 0 never fires. Same for `Unknown flag: $1` at line 155. The code doesn't deliver the property the comment promises.

### Testing
- [x] Test 28's 4 assertions are well-shaped:
  - Positive case (hooksPath set after run)
  - Activation log present on first run
  - Idempotent on re-run (no log)
  - Silent skip when hook file absent
- [x] Test uses real `git init` + real `git config` — exercises the actual path, not a mock
- [x] `bash -c "cd '$T28' && '$SCRIPT'"` correctly establishes cwd before the script's relative path checks fire
- [x] All 6 test suites green (211 total)
- [ ] **Test 28 only exercises the path where `GCP_PROJECT_ID` IS set.** It doesn't cover the "user runs script without env var" partial-run path that the Step 0 comment promises. Once Step 0 moves above the required-check (per the Required Change below), an additional test asserting that hook activation still fires when `GCP_PROJECT_ID` is unset would close the gap.

### Security
- [x] No new credential surface — `git config --local` only writes to the repo's `.git/config`
- [x] No injection vectors in the new shell paths (no user input flows into `git config`)
- [x] `--local` correctly scopes the change to this clone only

### Recommendation
**NO-GO** — one real correctness gap

The functional fix works for the happy path (user provides `GCP_PROJECT_ID`, script runs, hook activates). But the comment block claims a stronger property — "BEFORE any failable GCP call so even a partial provisioner run leaves the quality gate active" — that the placement doesn't deliver. The most common partial-run case (user forgets `GCP_PROJECT_ID`) bails at line 162 before Step 0 fires. Workshop attendees on day 1 are exactly the people who'll hit that path.

1. **Move the Step 0 block above the `GCP_PROJECT_ID` required-check** (currently lines 159-163). Best position: immediately after the `RETRY_MAX_ATTEMPTS` / `RETRY_MAX_SLEEP` defaults at line 84-85, BEFORE the argument-parsing loop. That ensures activation fires before:
   - The arg-parsing `exit 1` (line 155 `Unknown flag`)
   - The GCP_PROJECT_ID required-check `exit 1` (line 162)
   - Any GCP call

   The banner echo can stay where it is (line 165-173) since it's purely informational; Step 0 doesn't need to be visually adjacent to it. Update the comment to drop "between the banner and Step 1" if it's there, or just relocate without changing wording.

   Implementation note: `[[ -d .git ]]` and `[[ -f scripts/hooks/pre-push ]]` checks already gate the activation correctly — they're the only safety the early placement needs.

2. **Add an assertion to Test 28 that activation fires even when `GCP_PROJECT_ID` is unset.** This pins the property the comment claims:

   ```bash
   # Assertion 5: activation fires even before the GCP_PROJECT_ID check
   T28c=$(mktemp -d -t aflowtest-XXXX)
   git init -q -b main "$T28c"
   mkdir -p "$T28c/scripts/hooks"
   echo '#!/usr/bin/env bash' > "$T28c/scripts/hooks/pre-push"

   set +e
   # Note: NO GCP_PROJECT_ID set — script will exit 1, but Step 0 should
   # still have run.
   PATH="$T28/bin:$PATH" bash -c "cd '$T28c' && '$SCRIPT'" > "$T28c/run.log" 2>&1
   set -e

   hooks_path=$(git -C "$T28c" config --local --get core.hooksPath 2>/dev/null || echo "(unset)")
   assert_eq "scripts/hooks" "$hooks_path" "core.hooksPath set even when GCP_PROJECT_ID is unset"
   ```

   This is the regression guard for the property. Without it, a future refactor that re-introduces the placement bug ships green.

- The doctor.sh fix flag-rename is good, but consider also bumping the existing `pre-push exists + executable (WARN)` check at line 252-258 to a FAIL if you want full hook coverage. Currently a fork could have core.hooksPath set correctly, the pre-push file present but non-executable, and doctor only WARNs. Out of scope for this PR.
- The `[[ -d .git ]]` gate is correct for the standard case but technically misses sub-worktrees (`.git` is a file pointing at the gitdir there). `git rev-parse --is-inside-work-tree >/dev/null 2>&1` would handle both. Genuinely edge-case; current check is fine for the workshop audience but worth knowing.
- The GETTING-STARTED prose says "every fresh clone needs this command once" — true and clear. Worth a follow-up: have the bootstrap wizard (`/bootstrap-product`?) check for it and prompt, since the wizard runs once per clone too. Defer-able.

va-worker:
Pushed `50e9de3` addressing both required changes.

**1. Step 0 relocated above input validation.** Moved from its original spot (between the banner and Step 1, line 175) to immediately after the `retry_eventual_consistency` helper definition and before the arg-parsing loop (line 144). Activation now fires before:
- `Unknown flag` exit 1 (line 155)
- `GCP_PROJECT_ID is required` exit 1 (line 184)
- The banner echo
- All GCP calls

The comment is updated to reflect the new ordering.

**2. Test 28 gains Assertion 5 (3 sub-assertions).** Runs the provisioner WITHOUT `GCP_PROJECT_ID` set and verifies:
- `script exits 1` (precondition — confirms we're exercising the failure path, not the happy path)
- `core.hooksPath` is still set to `scripts/hooks`
- The `[hook] Activated` message was logged before the exit

This is the regression guard for Step 0's placement. A future refactor that regresses the ordering fails this test loudly.

All 6 test suites green (97 → 100 in the provision suite; 211 → 214 total). shellcheck clean. Ready for re-review.

va-reviewer:
Re-review (post-merge): Step 0 placement at line 144 confirmed above the `GCP_PROJECT_ID` required-check (line 184). Both required changes from the previous NO-GO landed correctly. CI green. Audit trail: GO.

PR #81, opened 2026-04-30
fix(app/workflow): preview deploys fail fast on missing DATABASE_URL; smoke test probes DB

### LINKED ISSUE #78: fix(app/workflow): preview deploys silently fall through to SQLite when DATABASE_URL is unset
## Problem Statement

When `NEON_API_KEY` is unset in a fork, `preview-deploy.yml`'s Neon branch step is skipped (correctly gated). The deploy then ships a Cloud Run revision with no `DATABASE_URL` env var. The container starts; pydantic-settings falls through to the dev-default `sqlite:///./dev.db`; the home route 500s on first request with `sqlite3.OperationalError: no such table: todo`.

The validator added in #63 only fires for `ENVIRONMENT=production`. Previews bypass it.

The smoke test only probes `/api/health`, which deliberately doesn't touch the DB. Result: deploy succeeds, smoke test passes, PR comment reads "Ready for testing", and the URL is broken — which a human only finds by clicking it.

Surfaced 2026-04-30 dry-run on `tck517/tck517-app` PR #1. Long-form analysis in `tck517/tck517-app:docs/UPSTREAM-HANDOFF.md` Issue #1.

**Effort Estimate:** S
**Priority:** P0 — every fork without Neon configured hits this on PR #1. Workshop-prep work assumed Neon would be configured cohort-wide; that assumption doesn't hold for fork-only previews and silently breaks them.

---

## A. Environment Context

- File: `app/config.py:37-58` — `_refuse_sqlite_in_production` validator gates on `environment != "production"`. Previews send `ENVIRONMENT=preview` and skip the guard.
- File: `tests/test_config.py` — has `prod+empty` and `prod+sqlite` cases; needs `preview+sqlite` and `preview+postgres`.
- File: `.github/workflows/preview-deploy.yml:193-219` — smoke test only probes `/api/health`. The handler at `app/api/health.py:17` deliberately doesn't touch the DB (so Neon cold-wake on every probe is avoided), so a totally broken DB still passes.
- File: `app/api/health.py` — currently has `/api/health` (no DB). Add `/api/health/db` that does `SELECT 1` for the smoke test to probe.

---

## B. Guardrails

- Local dev (`ENVIRONMENT=development`, default) MUST continue to work with the SQLite fallback. No regressions on `pytest` runs that don't set `ENVIRONMENT`.
- The validator must accept `test` as well — pytest runs that explicitly set `ENVIRONMENT=test` should not need a real Postgres. Treat `development` and `test` as the only "SQLite OK" environments.
- The new `/api/health/db` endpoint must be a single-query probe (`SELECT 1`), not a heavyweight check that pulls connection-pool metadata or runs migrations. Cheap on Neon cold-start.
- Do NOT add a separate workflow-level "Verify preview has a database URL" step (the handoff doc proposed this as Fix A). The validator alone catches both Neon-unset and Neon-set-but-broken paths, fails sooner (at container startup), and avoids two-places-to-keep-in-sync drift. The handoff doc misdiagnosed the workflow as the failure point — verified empirically that `if [ -n "..." ]` at `preview-deploy.yml:151` does correctly skip the append when Neon output is empty. The bug is purely on the app side.
- The smoke test addition must keep the existing `/api/health` probe (load-balancer health check pattern) and add the DB probe as a second step. Both must pass for the workflow to report green.

---

## C. Happy Path

**Fix B (validator generalization)** — single line change in `app/config.py`:

```diff
     @model_validator(mode="after")
-    def _refuse_sqlite_in_production(self) -> Self:
+    def _refuse_sqlite_outside_dev(self) -> Self:
         # Defense-in-depth: the dev-default sqlite URL must never reach
-        # a production runtime. If the secret-mounted DATABASE_URL is
-        # missing or the mount didn't override the default, fail at
-        # startup with a clear message rather than 500ing on the first
-        # DB query with `no such table: todo`. See #63 / #71.
-        if self.environment != "production":
+        # any non-dev runtime (production, preview, staging). If
+        # DATABASE_URL is missing or doesn't override the default, fail
+        # at startup with a clear message rather than 500ing on the
+        # first DB query with `no such table: todo`. See #63 / #71 /
+        # #<this-issue>.
+        if self.environment in ("development", "test"):
             return self
```

After this lands, a preview deploy without a real Postgres URL fails fast at `Settings()` construction with the existing error messages — Cloud Run logs show a `ValueError` at startup, not a downstream 500.

**Fix C (smoke test)** — add a DB probe.

```python
# app/api/health.py — new endpoint
@router.get("/api/health/db")
def health_db(session: Session = Depends(get_session)) -> dict[str, str]:
    session.exec(select(1)).one()
    return {"status": "ok"}
```

```yaml
# preview-deploy.yml — new probe after the existing /api/health loop
echo "Testing DB-backed route: $PREVIEW_URL/api/health/db"
HTTP_STATUS=$(curl -s -o /dev/null -w "%{http_code}" --max-time 15 "$PREVIEW_URL/api/health/db")
if [ "$HTTP_STATUS" != "200" ]; then
  echo "::error::DB health probe returned $HTTP_STATUS — DB or migrations may be broken"
  echo "result=failed" >> "$GITHUB_OUTPUT"
  exit 1
fi
```

After this lands, a preview that boots but has a broken DB connection (wrong password, missing migration, etc.) fails the smoke test loudly rather than passing on the no-DB `/api/health` route alone.

---

## D. Definition of Done

- [ ] `app/config.py` validator generalized: gates on `environment in ("development", "test")` instead of `environment != "production"`. Method renamed (`_refuse_sqlite_outside_dev`) and docstring updated.
- [ ] `tests/test_config.py` adds 2+ new cases:
  - `preview + sqlite://` → `ValueError`
  - `preview + valid postgres URL` → no raise (round-trips through field validator's psycopg3 normalization)
  - `staging + sqlite://` → `ValueError` (treats unfamiliar non-dev environment correctly)
- [ ] `app/api/health.py` adds `/api/health/db` endpoint that runs `SELECT 1`
- [ ] `tests/test_health.py` covers the new endpoint (200 with working DB; covers the imports/wiring even if it doesn't reproduce a broken-DB case)
- [ ] `.github/workflows/preview-deploy.yml` smoke test extended: keeps `/api/health` probe loop, adds a single `/api/health/db` probe after it. Both must pass for green.
- [ ] `pytest` clean; `mypy app/` clean; `ruff` clean
- [ ] Verified end-to-end:
  - Fork with no Neon configured opens PR → preview-deploy fails at smoke test (or at container startup; either is acceptable)
  - Fork with Neon configured opens PR → preview-deploy succeeds, both probes pass, URL serves FastAPI

---

## Why P0

Every fork without Neon configured ships a broken preview on PR #1. Workshop attendees who don't get Neon set up before opening their first PR see a successful deploy + green smoke test + broken URL — which is the worst-of-both-worlds failure mode the framework is supposed to prevent.

Fix is small (validator: 1 line; new endpoint: ~5 lines; workflow probe: ~5 lines).

---

## Note on the handoff doc's Fix A

The handoff doc proposed a third fix — a "Verify preview has a database URL" workflow step before deploy. After verifying the workflow logic at `preview-deploy.yml:151`, that fix is unnecessary: the existing `if [ -n "..." ]` guard does correctly skip the `DATABASE_URL=` append when Neon output is empty. The actual failure point is purely the app-side validator gap. Adding a workflow step would duplicate what the validator does and create two places to keep in sync.

va-worker:
## Summary

Closes #78. Two-layer defense for the 2026-04-30 dry-run failure mode: preview deploys without Neon configured were silently falling through to the dev-default `sqlite:///./dev.db`, booting cleanly, passing the no-DB `/api/health` smoke test, and 500ing on the home route with `no such table: todo`. Looked like a schema bug; was a config bug.

## Why

The validator added in #63 only fired for `ENVIRONMENT=production`. Previews send `ENVIRONMENT=preview` and bypass the guard. The smoke test at `preview-deploy.yml:193-219` only probes `/api/health` — which deliberately doesn't touch the DB — so a totally broken database connection still passes. PR comment reads "Smoke test: passed". Human opens the URL → 500.

Surfaced 2026-04-30 on `tck517/tck517-app` PR #1.

### Validator generalization (`app/config.py`)

Was gated on `environment != "production"`. Now gates on `environment in ("development", "test")` — so **preview, staging, and any other non-dev environment** fail fast at `Settings()` construction when `DATABASE_URL` is empty or sqlite-shaped.

```diff
-    def _refuse_sqlite_in_production(self) -> Self:
-        if self.environment != "production":
+    def _refuse_sqlite_outside_dev(self) -> Self:
+        if self.environment in ("development", "test"):
             return self
```

Error messages now name the actual environment ("DATABASE_URL is empty in preview" rather than always "in production") and call out the most likely cause for previews specifically (NEON_API_KEY unset so the Neon branch step was skipped).

### New endpoint (`app/api/health.py`)

```python
@router.get("/api/health/db")
def health_db(session: SessionDep) -> JSONResponse:
    """Single-query DB probe — `SELECT 1`."""
    session.execute(text("SELECT 1"))
    return JSONResponse({"status": "ok"})
```

Distinct from `/api/health`: catches the revision-up-but-DB-broken case. Uses the project's existing `SessionDep` Annotated alias pattern (matches `todos.py`) to keep the FastAPI dependency-injection idiom consistent and satisfy ruff `B008`. `session.execute(text(...))` is the right call because SQLModel's typed `session.exec()` only accepts SQLModel statements, not raw text.

### Smoke test restructure (`.github/workflows/preview-deploy.yml`)

Two-phase probe replacing the original single-phase loop:

- **Phase 1**: retry loop against `/api/health` (Cloud Run cold-start + Neon wake — original behavior, preserved)
- **Phase 2**: single-shot probe against `/api/health/db` — must return 200 for green

Both must pass.

## Tests

| File | Cases added |
|---|---|
| `tests/test_config.py` | `preview + sqlite → ValueError`; `preview + postgres → normalized + accepted`; `staging + sqlite → ValueError` (future-proofing for unfamiliar environments); `test + sqlite → accepted` (explicit allow case for pytest fixtures) |
| `tests/test_health.py` | `/api/health/db` returns 200 with a working session |

Total pytest: 17 → **22** (+5). ruff + format + mypy clean.

## What this PR does NOT include

The original handoff doc proposed a third fix — a "Verify preview has a database URL" workflow step before deploy. After verifying the workflow logic at `preview-deploy.yml:151`, that fix is unnecessary: the existing `if [ -n "..." ]` guard does correctly skip the `DATABASE_URL=` append when Neon output is empty. The actual failure point is purely the app-side validator gap. Adding a workflow step would duplicate what the validator does and create two places to keep in sync.

## Verification post-merge

- [ ] CI green
- [ ] After merge: open a PR on a fork without `NEON_API_KEY`. Confirm preview-deploy fails (either at container startup with the validator's `ValueError` in Cloud Run logs, or at the smoke test's Phase 2 — either is acceptable; the failure is loud either way).
- [ ] After merge: open a PR on a fork WITH Neon configured. Confirm preview-deploy succeeds, both phases pass, URL serves FastAPI.

### DIFF
--- a/.github/workflows/preview-deploy.yml
+++ b/.github/workflows/preview-deploy.yml
@@ -202,21 +202,45 @@ jobs:
             exit 1
           fi
 
-          echo "Testing: $PREVIEW_URL/api/health"
+          # Phase 1: app readiness — /api/health does NOT touch the DB,
+          # so a passing probe means the container booted and is serving
+          # HTTP. Retry loop covers Cloud Run cold-start + Neon wake.
+          echo "Phase 1: $PREVIEW_URL/api/health"
 
+          health_passed=false
           for ATTEMPT in $(seq 1 10); do
             HTTP_STATUS=$(curl -s -o /dev/null -w "%{http_code}" --max-time 15 "$PREVIEW_URL/api/health" || echo "000")
             if [ "$HTTP_STATUS" = "200" ]; then
               echo "Health check passed"
-              echo "result=passed" >> "$GITHUB_OUTPUT"
-              exit 0
+              health_passed=true
+              break
             fi
             echo "Attempt $ATTEMPT: HTTP $HTTP_STATUS, retrying in 6s..."
             sleep 6
           done
 
-          echo "result=failed" >> "$GITHUB_OUTPUT"
-          exit 1
+          if [ "$health_passed" != "true" ]; then
+            echo "::error::Container readiness probe failed after 10 attempts"
+            echo "result=failed" >> "$GITHUB_OUTPUT"
+            exit 1
+          fi
+
+          # Phase 2: DB reachability — /api/health/db runs `SELECT 1`.
+          # Distinct from /api/health: catches the failure mode where
+          # the container is up but DATABASE_URL is empty/broken (added
+          # in #78 after a 2026-04-30 dry-run shipped a green-CI preview
+          # with no DB and an opaque 500 on the home route).
+          echo "Phase 2: $PREVIEW_URL/api/health/db"
+
+          HTTP_STATUS=$(curl -s -o /dev/null -w "%{http_code}" --max-time 15 "$PREVIEW_URL/api/health/db" || echo "000")
+          if [ "$HTTP_STATUS" != "200" ]; then
+            echo "::error::DB health probe returned $HTTP_STATUS — DB connection or migrations may be broken"
+            echo "result=failed" >> "$GITHUB_OUTPUT"
+            exit 1
+          fi
+
+          echo "DB health check passed"
+          echo "result=passed" >> "$GITHUB_OUTPUT"
 
       # ── Comment on the PR with preview info ───────────────────────────
 
--- a/app/api/health.py
+++ b/app/api/health.py
@@ -1,13 +1,26 @@
-"""Health check endpoint.
+"""Health check endpoints.
 
-Cloud Run uses this for container readiness checks. The preview-deploy
-workflow also hits this as the smoke test.
+`/api/health` — fast probe used by Cloud Run readiness checks. Does NOT
+touch the database (Neon cold-wake on every probe is expensive).
+
+`/api/health/db` — DB-backed probe used by the preview-deploy smoke test
+to verify the deploy actually has a working database connection. Added
+in #78 because preview deploys without Neon configured were passing
+the no-DB `/api/health` probe and shipping a broken URL.
 """
 
-from fastapi import APIRouter
+from typing import Annotated
+
+from fastapi import APIRouter, Depends
 from fastapi.responses import JSONResponse
+from sqlalchemy import text
+from sqlmodel import Session
+
+from app.db import get_session
 
 router = APIRouter()
+
+SessionDep = Annotated[Session, Depends(get_session)]
 
 
 @router.get("/api/health")
@@ -17,4 +30,19 @@ async def health() -> JSONResponse:
     Deliberately does NOT touch the database — health checks must be
     fast and not wake up the Neon compute endpoint on every probe.
     """
+    return JSONResponse({"status": "ok"})
+
+
+@router.get("/api/health/db")
+def health_db(session: SessionDep) -> JSONResponse:
+    """Single-query DB probe — `SELECT 1`.
+
+    Used by preview-deploy.yml to validate that the deployed revision
+    can reach the database. Cheap (single round-trip, no schema), but
+    distinct from `/api/health`: a revision with `DATABASE_URL=""` or
+    a misconfigured Neon branch fails this and passes `/api/health`.
+    Uses SQLAlchemy's `session.execute(text(...))` because SQLModel's
+    typed `session.exec()` only accepts SQLModel statements.
+    """
+    session.execute(text("SELECT 1"))
     return JSONResponse({"status": "ok"})
--- a/app/config.py
+++ b/app/config.py
@@ -35,26 +35,31 @@ class Settings(BaseSettings):
         return v
 
     @model_validator(mode="after")
-    def _refuse_sqlite_in_production(self) -> Self:
+    def _refuse_sqlite_outside_dev(self) -> Self:
         # Defense-in-depth: the dev-default sqlite URL must never reach
-        # a production runtime. If the secret-mounted DATABASE_URL is
-        # missing or the mount didn't override the default, fail at
-        # startup with a clear message rather than 500ing on the first
-        # DB query with `no such table: todo`. See #63 / #71.
-        if self.environment != "production":
+        # any non-dev runtime (production, preview, staging). If
+        # DATABASE_URL is missing or doesn't override the default, fail
+        # at startup with a clear message rather than 500ing on the
+        # first DB query with `no such table: todo`. Originally only
+        # gated `production`; #78 generalized after a 2026-04-30 dry-run
+        # showed previews silently fell through to SQLite when Neon was
+        # unconfigured. See #63 / #71 / #78.
+        if self.environment in ("development", "test"):
             return self
         if not self.database_url:
             raise ValueError(
-                "DATABASE_URL is empty in production. "
-                "Check that the Secret Manager secret 'database-url' is "
-                "mounted in deploy.yml and the Cloud Run service has "
-                "secretAccessor on it."
+                f"DATABASE_URL is empty in {self.environment}. "
+                "Check that DATABASE_URL is set on the Cloud Run revision "
+                "(production reads from PRODUCTION_DATABASE_URL secret; "
+                "preview reads from the Neon branch action output)."
             )
         if self.database_url.startswith("sqlite"):
             raise ValueError(
-                f"DATABASE_URL is SQLite in production: {self.database_url!r}. "
-                "Production must use Postgres. Likely cause: the secret-mounted "
-                "DATABASE_URL env var didn't override the dev default."
+                f"DATABASE_URL is SQLite in {self.environment}: {self.database_url!r}. "
+                "Non-dev environments must use Postgres. Likely cause: the "
+                "DATABASE_URL env var didn't override the dev default — for "
+                "previews, NEON_API_KEY may be unconfigured so the Neon branch "
+                "step was skipped."
             )
         return self
 
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -98,3 +98,54 @@ def test_production_postgres_url_passes(monkeypatch: pytest.MonkeyPatch) -> None
 
     settings = Settings()
     assert settings.database_url == "postgresql+psycopg://u:p@h.neon.tech/db"
+
+
+def test_preview_sqlite_database_url_fails(monkeypatch: pytest.MonkeyPatch) -> None:
+    # The 2026-04-30 dry-run failure mode: Neon unconfigured, preview
+    # falls through to the dev-default `sqlite:///./dev.db`, app boots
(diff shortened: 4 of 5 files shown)


va-reviewer:
## PR Review — #81

### Requirements
- [x] Acceptance criteria from #78 are met (validator generalized; new `/api/health/db`; smoke test extended; tests cover the new paths)
- [x] Feature works end-to-end as described — independently verified the failure chain: misconfigured `DATABASE_URL` → `Settings()` raises at `app/db.py:14` import time → uvicorn never starts → Cloud Run logs show `ValueError` → smoke test Phase 1 retry loop times out within ~60s
- [x] No scope creep — three precise changes (validator gate, new endpoint, smoke probe) plus their tests. The original handoff doc's "Fix A" workflow-level pre-deploy check is correctly omitted with a clear rationale

### Code Quality
- [x] Validator gate inversion (`environment != "production"` → `environment in ("development", "test")`) is the right shape. Allow-list semantics (explicit dev environments) beats deny-list (every non-prod treated as prod) — adding new environments to the allow list is a one-line change, while adding them to a deny list would require remembering every site that mentions "production"
- [x] Method renamed `_refuse_sqlite_in_production` → `_refuse_sqlite_outside_dev` — name now matches behavior; less misleading than leaving the old name on a generalized validator
- [x] Error messages now interpolate `self.environment` so the message says "DATABASE_URL is empty in preview" / "in staging" instead of always claiming "in production". Cause-hint for previews specifically calls out NEON_API_KEY — high-signal for the workshop audience
- [x] `/api/health/db` correctly uses `session.execute(text("SELECT 1"))` rather than `session.exec()`. Verified independently against SQLModel — the typed `exec()` only accepts SQLModel statements; `execute()` is the right escape hatch for raw SQL. Inline comment explains the choice
- [x] Endpoint adopts the project's `SessionDep = Annotated[Session, Depends(get_session)]` pattern from `todos.py:25`. Consistent style + dodges ruff B008
- [x] Smoke test refactor preserves the existing `/api/health` retry-loop behavior (Cloud Run cold-start + Neon wake) and adds Phase 2 as a single-shot probe — correct asymmetry: readiness needs retries, the post-readiness DB check does not
- [x] `::error::` annotations on both Phase 1 and Phase 2 failures land in the GitHub Actions UI, not just stdout; future-proofs the failure visibility

### Testing
- [x] 5 new test cases cover the contract precisely:
  - `preview + sqlite` → ValueError (the 2026-04-30 failure mode itself)
  - `preview + postgres` → normalized + accepted (regression guard for the happy path)
  - `staging + sqlite` → ValueError (future-proofing for unfamiliar environments)
  - `test + sqlite` → accepted (explicit allow case, prevents the validator from breaking pytest fixtures that set `ENVIRONMENT=test`)
- [x] `match=` on each `pytest.raises` checks the specific user-visible message text (e.g., `"DATABASE_URL is SQLite in preview"`) — meaningful regression guard for the helpful environment naming
- [x] `/api/health/db` happy path test added in `test_health.py`. The existing test client fixture wires a real (in-memory SQLite) session, so the test exercises the actual `SELECT 1` round-trip
- [x] All CI checks green; ran the full pytest suite locally on the PR head (16 config+health cases, all green, 0.04s)

### Security
- [x] No new credential surface — `/api/health/db` reuses the existing `get_session` dependency, no new connection paths
- [x] No information disclosure: `/api/health/db` returns `{"status": "ok"}` on success and 500 (FastAPI default for unhandled exceptions) on failure. Doesn't leak DB internals to the public probe surface
- [x] `text("SELECT 1")` is a hardcoded literal — no user input flows into it; no SQL injection surface
- [x] Error message in `_refuse_sqlite_outside_dev` includes `{self.database_url!r}` for the sqlite case. For the realistic prod failure (the dev-default `sqlite:///./dev.db`) this is fine — no credentials. Same edge-case caveat as #73's review applies (sqlite URLs don't carry credentials), still acceptable

### Recommendation
**GO**

Defense-in-depth done right. Validator catches the failure mode at the earliest possible point (import time, before uvicorn binds); smoke test catches the orthogonal "container up but DB unreachable" case that the validator can't see. Allow-list semantics will scale cleanly to staging/canary environments without code changes. CI green; full suite locally green; code matches the project's existing FastAPI dependency-injection idiom.

The PR description's explicit "what this PR does NOT include" callout (handoff doc's Fix A) demonstrates good scope discipline — the rationale ("validator alone catches both Neon-unset and Neon-set-but-broken paths, fails sooner, avoids two-places-to-keep-in-sync drift") matches what an independent review would conclude.

None.

- The pre-existing `_settings = get_settings()` at `app/db.py:14` constructs `Settings()` at module import time — that's why the validator fires before uvicorn binds, which is a feature here. But it also means a single test that imports `app.db` with bad env vars will crash collection. The conftest's `monkeypatch` fixture-scoped overrides probably handle this, but worth a future ticket to verify (and possibly defer construction to `get_session()`'s first call) if you ever start running prod-shaped env values in test environments.
- Consider an `/api/health/migrations` endpoint that asserts the schema is at the expected Alembic revision. Distinct from `/api/health/db` (DB reachable) and `/api/health` (process up). Catches the "DB reachable but migrations didn't run" case which is the OTHER way to ship a broken preview. Out of scope for this PR; worth a follow-up ticket.
- Pattern entries #31 and #32 in `PATTERN-LIBRARY.md` could use a follow-up entry capturing this allow-list-vs-deny-list lesson — "when extending validators across environments, prefer allow-listing the safe environments." Pure docs; tracking-only.
- The smoke test's `health_passed` boolean is correct but could be more idiomatic as `if [ "$health_passed" = "true" ]` rather than `!= "true"` (positive checks read more naturally). Pure style; no behavior difference.

PR #96, opened 2026-04-30
feat(neon): document branch model + detect parent-branch collision (closes #90)

### LINKED ISSUE #90: feat(neon): document branch model + detect parent-branch collision in provisioner
## Problem Statement

The framework's Neon branch model is undocumented and has a silent-cross-contamination footgun for the workshop's most-common deployment shape (one shared Neon project, N attendees).

Specifically, `scripts/provision-gcp-project.sh` Step 5.7 currently treats a branch-name collision as a benign no-op:

```bash
elif [[ "$http_code" == "409" ]]; then
  echo "[neon] branch '$NEON_BRANCH_NAME' already exists; reusing"
```

(`scripts/provision-gcp-project.sh:656-657`)

When two attendees share a Neon project AND have the same handle in the roster CSV (`alice`), the second provisioner run **silently mounts the second attendee's `database-url` Secret Manager secret to the first attendee's Neon branch**. Same with `NEON_PARENT_BRANCH` on the GitHub side: the second attendee's previews inherit the first attendee's seed data. No error, no warning, no diagnostic — looks identical to a fresh setup until someone notices their data appearing in someone else's app.

Adjacent gap: `preview-deploy.yml:80` defaults `parent` to the literal string `'main'`, not the project's actual default branch. A Neon project that ships with `production` (or any other name) as its default silently fails or creates a duplicate branch with no warning.

Adjacent gap: branches in the Neon console are named after roster handles. When the roster includes bot-account-style handles (e.g., `va-worker`, `va-reviewer` if a facilitator is dogfooding the bot accounts as test attendees), those Neon branches are indistinguishable from leftover artifacts to anyone who didn't write the roster. **The framework should document the branch list a user should expect to see, so they can identify what's intentional vs what isn't.**

Surfaced 2026-04-30 in `tck517/tck517-app:docs/UPSTREAM-HANDOFF.md` Issue #5. The reporter saw three branches in their Neon console (`production`, `tck517`, `va-worker`) — all three legitimate (the third was a `va-worker` row in the roster), but the framework's docs gave no way to confirm that without spelunking the roster CSV.

**Effort Estimate:** S
**Priority:** P2 — no production damage observed, but the silent-cross-contamination case is a real workshop scenario that produces confusing results without any visible error.

---

## A. Environment Context

- File: `docs/PLATFORM-GUIDE.md` — closest thing to Neon docs today (lines 122-126 in the secrets table). Needs a dedicated "Neon branch model" section.
- File: `scripts/provision-gcp-project.sh:656-657` — the silent-reuse-on-409 logic.
- File: `scripts/provision-workshop-roster.sh:155-156` — defaults `neon_branch` to `handle`. Collision possible when two roster rows share a handle, OR if a roster row's handle matches an already-existing branch from a prior cohort.
- File: `.github/workflows/preview-deploy.yml:80` — `parent: ${{ secrets.NEON_PARENT_BRANCH || 'main' }}`. Hardcoded fallback.
- File: `.github/workflows/preview-cleanup.yml` — uses `neondatabase/delete-branch-action@v3`. Cleanup is shared-project-wide.

---

## B. Guardrails

- **No defect rewrite.** Don't change `preview-deploy.yml`'s `pr-{N}` naming convention — PR numbers are unique-per-repo, so cross-fork preview branches do not collide. That part of the model is correct.
- **Don't break legitimate intentional sharing.** Two attendees who genuinely want to share a parent branch (e.g., paired attendees collaborating on the same dataset) need an opt-in path — `--force-shared-parent` flag or similar. Default behavior changes from "silent reuse" to "fail with actionable message"; the explicit-share flag preserves the silent-reuse semantics.
- **Cross-cohort collisions are a real case.** A facilitator running the workshop a second time with a fresh roster but the same Neon project will hit collisions on every attendee who shared a handle with a prior cohort. The fail-loud message should mention this case.
- **Docs change is the most valuable single output.** Provisioner detection (Fix B) is a smaller win than the docs section (Fix A). If we can ship only one, ship Fix A.
- **`production` vs `main` mismatch is a separate, smaller fix.** Bundle with this PR only if cheap; otherwise file as a follow-up.

---

### Fix A — `docs/PLATFORM-GUIDE.md` "Neon branch model" section

New top-level section explaining:

1. **Branch namespace** — Neon branches are unique within a project. Across projects, names can repeat freely.
2. **Branch types in this framework**:
   - `main` — the project's default. Production app's pooled connection points here. **Must be named `main`** for the framework's `preview-deploy.yml:80` fallback to work; if your Neon project uses a different default name (e.g., `production`), set `NEON_PARENT_BRANCH` explicitly to that name.
   - `<roster-handle>` — long-lived seed branches, one per roster row. Handle defaults to the roster's `handle` column. Created once during onboarding by Step 5.7 of `provision-gcp-project.sh`. Never auto-deleted.
   - `pr-{N}` — short-lived. Auto-created by `preview-deploy.yml`, auto-deleted by `preview-cleanup.yml`. PR numbers are unique-per-repo, so these do not collide across forks.
3. **What `NEON_PARENT_BRANCH` does** — determines what schema and seed data each PR's preview branch inherits. Default `main` means "inherit from production schema, no seed data." Setting it to an attendee's name means "inherit from my seeded dev data."
4. **Workshop multi-tenancy model** — facilitators typically provision **one** shared Neon project per cohort (cost-efficient). All attendees see each other's branches in the Neon console. **Branch deletion is project-wide** — if cleanup fires for `pr-3`, it's gone for the entire project. PR-number uniqueness across forks prevents this from happening in normal use, but be aware if you ever construct PRs with manually-set numbers.
5. **What you should expect to see in the console** — given a roster of N attendees, expect: 1× `main` (the default), N× `<handle>` (one per row), 0+× `pr-{N}` (one per currently-open PR across all forks). Anything else is a leftover and can be inspected against the roster CSV.
6. **Collision behavior** — two roster rows with the same handle, OR a single row whose handle matches a branch from a prior cohort, results in the second provisioner run **failing** (post Fix B) with a clear message. To intentionally share a parent (paired collaboration), pass `--force-shared-parent` to the wrapper.

### Fix B — Detect parent-branch collision in the provisioner

Replace the silent-reuse path at `provision-gcp-project.sh:656-657` with:

```bash
elif [[ "$http_code" == "409" ]]; then
  if [[ "${NEON_FORCE_SHARED_PARENT:-false}" == "true" ]]; then
    echo "[neon] branch '$NEON_BRANCH_NAME' exists; reusing per --force-shared-parent"
    # ... existing lookup logic ...
  else
    echo "ERROR: Neon branch '$NEON_BRANCH_NAME' already exists in project '$NEON_PROJECT_ID'." >&2
    echo "  This may mean: another roster row already used this handle," >&2
    echo "  OR a prior cohort left a branch with this name." >&2
    echo "  Choose a different handle, OR set NEON_FORCE_SHARED_PARENT=true" >&2
    echo "  if you intentionally want to share the parent branch (paired" >&2
    echo "  collaboration, or re-running the same cohort)." >&2
    exit 1
  fi
fi
```

Add a `--force-shared-parent` flag to `scripts/provision-workshop-roster.sh` that propagates `NEON_FORCE_SHARED_PARENT=true` to the inner script.

### Fix C — Default-branch-name mismatch

Smallest fix: update the comment in `preview-deploy.yml:78-80` to explicitly call out that `'main'` is hardcoded and that Neon projects whose default branch is named differently must set `NEON_PARENT_BRANCH` explicitly. Add a one-liner to PLATFORM-GUIDE.md.

A more robust alternative (query the Neon API for the actual default branch name) is out of scope here — the static fallback works for every Neon-defaults shape we ship today.

---

## D. Definition of Done

- [ ] `docs/PLATFORM-GUIDE.md` adds a "Neon branch model" section covering namespace, branch types, workshop multi-tenancy, collision behavior, the `main`-naming requirement, AND the "what you should expect to see in the console" guidance for distinguishing legitimate roster-driven branches from leftovers
- [ ] `scripts/provision-gcp-project.sh:656-657` replaced with the fail-or-force logic above (with the prior-cohort hint in the error message)
- [ ] `scripts/provision-workshop-roster.sh` accepts `--force-shared-parent` and propagates as `NEON_FORCE_SHARED_PARENT=true`
- [ ] `scripts/provision-gcp-project.test.sh` extended: assert that 409 without the env var fails the script with a non-zero exit and the actionable message; assert that 409 WITH the env var preserves the existing silent-reuse behavior
- [ ] `preview-deploy.yml:78-80` comment updated to call out the `'main'`-hardcoded fallback
- [ ] All 6 test suites green
- [ ] `markdownlint` clean

---

## Why P2

Workshop is in 6 days; no defect has been observed at scale yet. But two attendees with the same handle is a realistic case (especially if facilitators don't pre-validate the roster, or a facilitator re-runs the workshop with a fresh roster against a still-populated Neon project). The current behavior produces silent cross-contamination that is genuinely confusing to debug. The docs change alone is a meaningful improvement; the provisioner detection adds a hard guard. Worth shipping before the workshop runs at full attendance.

Bundle with the existing solo-mode tickets (#82-#89) if there's bandwidth; otherwise this is the next P2 to pick up after the P0/P1s ship.

va-worker:
## Summary

Closes #90. Closes a silent-cross-contamination footgun in the Neon-branch flow + adds the canonical Neon Branch Model docs section.

## Why

Two real workshop scenarios collided with the framework's pre-#90 silent-reuse-on-409 behavior in `provision-gcp-project.sh:656-657`:

1. **Two attendees with the same handle** in the roster CSV: the second provisioner run silently mounted the second attendee's `database-url` Secret Manager secret to the first attendee's existing Neon branch. No error.
2. **Re-running a cohort against an already-populated Neon project** (common when a facilitator runs the workshop a second time with a fresh roster): every attendee whose handle matched a leftover branch from the prior cohort hit the same silent-reuse path.

Surfaced in `tck517/tck517-app:docs/UPSTREAM-HANDOFF.md` Issue #5.

### Fix B — Collision detection in provisioner

Replace silent-reuse path at `scripts/provision-gcp-project.sh:656` with fail-or-force gate:

```diff
   elif [[ "$http_code" == "409" ]]; then
-    echo "[neon] branch '$NEON_BRANCH_NAME' already exists; reusing"
+    if [[ "${NEON_FORCE_SHARED_PARENT:-false}" != "true" ]]; then
+      echo "ERROR: Neon branch '$NEON_BRANCH_NAME' already exists in project '$NEON_PROJECT_ID'." >&2
+      echo "  This may mean: another roster row already used this handle," >&2
+      echo "  OR a prior cohort left a branch with this name." >&2
+      echo "  Choose a different handle, OR set NEON_FORCE_SHARED_PARENT=true" >&2
+      echo "  if you intentionally want to share the parent branch (paired" >&2
+      echo "  collaboration, or re-running the same cohort)." >&2
+      rm -f "$create_response_file"
+      exit 1
+    fi
+    echo "[neon] branch '$NEON_BRANCH_NAME' already exists; reusing per NEON_FORCE_SHARED_PARENT=true"
```

The escape-hatch path preserves the pre-#90 silent-reuse behavior verbatim — paired collaboration and re-running existing cohorts continue to work, just opt-in.

### Fix A — `docs/PLATFORM-GUIDE.md` Neon Branch Model section

New top-level section between Workshop Lifecycle and Daily Operations, covering:
- Branch namespace (unique within project)
- Three framework branch types (`main`, `<roster-handle>`, `pr-{N}`) with owners + lifetimes
- What `NEON_PARENT_BRANCH` does
- Workshop multi-tenancy model
- **What you should expect to see in the console** — given a roster of N, expect 1× `main` + N× `<handle>` + 0+× `pr-{N}`. Anything else is checkable against `roster.csv`. Generalizes the `va-worker`-confusion case from the downstream report (the user saw a Neon branch named after a bot, didn't know if it was legitimate; turned out to be a roster row with that handle).
- Collision behavior with the new fail-fast guard

### Fix C — `preview-deploy.yml` comment update

Inline comment at `parent: ${{ secrets.NEON_PARENT_BRANCH || 'main' }}` now explicitly calls out that `'main'` is a literal-string fallback, not "the project's actual default branch." Projects with a different default name must set `NEON_PARENT_BRANCH` explicitly.

### Wrapper change — `--force-shared-parent` flag

`scripts/provision-workshop-roster.sh` accepts `--force-shared-parent` and propagates as `NEON_FORCE_SHARED_PARENT=true`. Wrapper arg parsing rewritten to support flags alongside the positional roster CSV; preserves the strict "exactly one positional arg" check.

```bash
# Default: fail fast on existing branch
bash scripts/provision-workshop-roster.sh roster.csv

# Intentional shared parent
bash scripts/provision-workshop-roster.sh roster.csv --force-shared-parent
```

## Tests

| Test | Coverage |
|---|---|
| 17 (updated) | Existing idempotent-rerun test now passes `--force-shared-parent`; asserts the new log line names the flag |
| 29 (new) | Neon 409 + force unset → exit 1 with actionable message, names escape hatch, **does NOT write database-url secret** (cross-contam guard) |
| 30 (new) | Neon 409 + force=true → silent reuse, no error, proceeds to database-url secret handling |

| Suite | Before | After |
|---|---|---|
| `provision-gcp-project.test.sh` | 100 | **106** |
| `provision-workshop-roster.test.sh` | 32 | 32 |
| `setup-solo-mode.test.sh` | 25 | 25 |
| `diagnose-cloudrun.test.sh` | 35 | 35 |
| `workshop-setup.test.sh` | 21 | 21 |
| `workshop-teardown.test.sh` | 18 | 18 |
| `ensure-github-account.test.sh` | 8 | 8 |
| **Total** | 239 | **245** |

shellcheck + markdownlint + actionlint + pytest all clean.

## Verification post-merge

- [ ] CI green
- [ ] On a re-run cohort: confirm provisioner exits 1 with the actionable message when a roster handle collides with a leftover branch
- [ ] On a paired-collaboration row: confirm `--force-shared-parent` enables the silent-reuse path with the new explicit log line
- [ ] PLATFORM-GUIDE.md "Neon Branch Model" section renders cleanly on GitHub

### DIFF
--- a/.github/workflows/preview-deploy.yml
+++ b/.github/workflows/preview-deploy.yml
@@ -75,7 +75,14 @@ jobs:
           branch_name: pr-${{ env.PR_NUMBER }}
           # Workshop forks set NEON_PARENT_BRANCH to the attendee's branch name
           # (e.g. "alice") so per-PR previews inherit that attendee's schema.
-          # Non-workshop forks leave it unset and fall back to "main".
+          # Non-workshop forks leave it unset and fall back to the literal
+          # string "main" — which assumes the Neon project's default branch
+          # is named "main". Projects that were created with a different
+          # default name (e.g. "production") MUST set NEON_PARENT_BRANCH
+          # explicitly per fork; the literal-string fallback below does not
+          # query the project's actual default and will silently fail or
+          # create a sibling. See docs/PLATFORM-GUIDE.md "Neon Branch Model"
+          # for the canonical naming requirements (#90).
           # `||` is fine inside `with:` — only `if:` rejects `secrets.*`.
           parent: ${{ secrets.NEON_PARENT_BRANCH || 'main' }}
           username: ${{ vars.NEON_DB_USER || 'neondb_owner' }}
--- a/docs/PLATFORM-GUIDE.md
+++ b/docs/PLATFORM-GUIDE.md
@@ -629,6 +629,109 @@ gitignored. They contain participant emails — never commit either.
 
 ---
 
+## Neon Branch Model
+
+Neon branches are how the framework gives each attendee an isolated
+database without provisioning a separate Neon project per person. This
+section documents the framework's branch types, the workshop
+multi-tenancy model, and the collision behavior — read it once before
+running a cohort, especially if you're reusing a Neon project across
+cohorts.
+
+### Branch namespace
+
+Neon branches are unique within a project. Across projects, names can
+repeat freely. The framework uses one Neon project per cohort by
+default.
+
+### Branch types
+
+The framework creates and consumes three categories of Neon branch:
+
+| Type | Owner | Lifetime | Created by |
+|---|---|---|---|
+| `main` | Project default | Permanent | Neon (at project creation) |
+| `<roster-handle>` | One per roster row | Long-lived | `provision-gcp-project.sh` Step 5.7 |
+| `pr-{N}` | One per open PR | Short-lived | `preview-deploy.yml` |
+
+- **`main`** — the project's default branch. The production app's
+  pooled connection points here. **Must be named `main` literally.**
+  `preview-deploy.yml:80` falls back to the literal string `'main'`
+  when `NEON_PARENT_BRANCH` is unset; if your Neon project's default
+  branch is named differently (`production`, etc.), set
+  `NEON_PARENT_BRANCH` explicitly per fork or expect the parent
+  fallback to silently fail.
+- **`<roster-handle>`** — long-lived seed branches, one per roster row.
+  Branch name defaults to the CSV's `handle` column. Created once
+  during onboarding by Step 5.7 of `provision-gcp-project.sh`. Never
+  auto-deleted — these survive across cohorts unless a facilitator
+  manually removes them.
+- **`pr-{N}`** — short-lived. Auto-created by `preview-deploy.yml`
+  on PR open, auto-deleted by `preview-cleanup.yml` on PR close. PR
+  numbers are unique-per-repo, so cross-fork preview branches do
+  not collide.
+
+### What `NEON_PARENT_BRANCH` does
+
+Determines what schema and seed data each PR's preview branch
+inherits. The default `main` means "inherit from production schema,
+no seed data." Setting it to an attendee's roster handle means
+"inherit from my seeded dev data." `provision-gcp-project.sh` Step 7
+auto-pushes this as a GitHub Actions secret on the attendee's fork.
+
+### Workshop multi-tenancy model
+
+Facilitators typically provision **one** shared Neon project per
+cohort (cost-efficient). Three implications:
+
+1. **All attendees see each other's branches in the Neon console.**
+   This is normal; not a privacy issue (branch names are roster
+   handles, not anything sensitive).
+2. **Branch deletion is project-wide.** If `preview-cleanup.yml`
+   deletes `pr-3`, it's gone for the entire project, not just the
+   triggering fork. PR-number uniqueness across forks prevents this
+   from causing real harm in normal use, but be aware.
+3. **What you should expect to see in the console.** Given a roster
+   of N attendees, expect: 1× `main` (the default), N× `<handle>`
+   (one per row), 0+× `pr-{N}` (one per open PR across all forks).
+   Anything else is checkable against `roster.csv` — if a console
+   branch isn't in the roster, it's a leftover.
+
+### Collision behavior (since #90)
+
+Two roster rows with the same handle, OR a single row whose handle
+matches a branch left over from a prior cohort, results in the
+provisioner **failing fast** with an actionable message:
+
+```text
+ERROR: Neon branch 'alice' already exists in project 'dark-grass-...'.
+  This may mean: another roster row already used this handle,
+  OR a prior cohort left a branch with this name.
+  Choose a different handle, OR set NEON_FORCE_SHARED_PARENT=true
+  if you intentionally want to share the parent branch (paired
+  collaboration, or re-running the same cohort).
+```
+
+To intentionally share a parent (paired collaboration, or
+re-running an existing cohort against the still-populated Neon
+project), set `NEON_FORCE_SHARED_PARENT=true` in the environment
+when running the provisioner. The wrapper accepts
+`--force-shared-parent` and propagates it.
+
+```bash
+# Default: fail fast on existing branch
+bash scripts/provision-workshop-roster.sh roster.csv
+
+# Intentional shared parent
+bash scripts/provision-workshop-roster.sh roster.csv --force-shared-parent
+```
+
+The previous behavior (silent reuse on 409) is preserved when the
+flag is set, so existing automation with intentional collisions
+keeps working — just turn the flag on explicitly.
+
+---
+
 ## Daily Operations
 
 ### Viewing Logs
--- a/scripts/provision-gcp-project.sh
+++ b/scripts/provision-gcp-project.sh
@@ -654,7 +654,29 @@ sys.exit('no default branch found in project response')
     branch_id="$(python3 -c "import json,sys; print(json.load(open('$create_response_file'))['branch']['id'])")"
     echo "[neon] branch created (id=$branch_id)"
   elif [[ "$http_code" == "409" ]]; then
-    echo "[neon] branch '$NEON_BRANCH_NAME' already exists; reusing"
+    # 409: a branch with this name already exists in the project.
+    #
+    # Default behavior (since #90): fail loud. Two roster rows with
+    # the same handle, OR a roster row whose handle matches a branch
+    # left over from a prior cohort, would otherwise silently mount
+    # this attendee's database-url Secret Manager secret to the
+    # OTHER attendee's Neon branch — invisible cross-contamination.
+    #
+    # NEON_FORCE_SHARED_PARENT=true preserves the original silent-reuse
+    # behavior for legitimate cases: paired attendees collaborating on
+    # the same dataset, or re-running the same cohort against an
+    # already-populated Neon project. Set explicitly; never default on.
+    if [[ "${NEON_FORCE_SHARED_PARENT:-false}" != "true" ]]; then
+      echo "ERROR: Neon branch '$NEON_BRANCH_NAME' already exists in project '$NEON_PROJECT_ID'." >&2
+      echo "  This may mean: another roster row already used this handle," >&2
+      echo "  OR a prior cohort left a branch with this name." >&2
+      echo "  Choose a different handle, OR set NEON_FORCE_SHARED_PARENT=true" >&2
+      echo "  if you intentionally want to share the parent branch (paired" >&2
+      echo "  collaboration, or re-running the same cohort)." >&2
+      rm -f "$create_response_file"
+      exit 1
+    fi
+    echo "[neon] branch '$NEON_BRANCH_NAME' already exists; reusing per NEON_FORCE_SHARED_PARENT=true"
     # Look up the existing branch's ID by name.
     branch_id="$(echo "$branches_json" | python3 -c "
 import json, sys
--- a/scripts/provision-gcp-project.test.sh
+++ b/scripts/provision-gcp-project.test.sh
@@ -1184,6 +1184,8 @@ EOF
   # Use ${var-default} (no colon) so an explicit empty string passed by
   # the caller — like test 15's "" for NEON_API_KEY — stays empty.
   # ${var:-default} would substitute on empty, which we don't want here.
+  # 5th arg is NEON_FORCE_SHARED_PARENT (default false; pass "true" for
+  # collision-reuse tests in the post-#90 world).
   set +e
   PATH="$tmp/bin:$PATH" \
     GCP_PROJECT_ID="af-step57-test" \
@@ -1191,6 +1193,7 @@ EOF
     NEON_API_KEY="${2-fake-api-key}" \
     NEON_PROJECT_ID="${3-fake-project-id}" \
     NEON_BRANCH_NAME="${4-alice}" \
+    NEON_FORCE_SHARED_PARENT="${5-false}" \
     "$SCRIPT" --create-project > "$tmp/stdout.log" 2>&1
   set -e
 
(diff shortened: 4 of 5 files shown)


va-reviewer:
## PR Review — #96

### Requirements
- [x] Acceptance criteria from #90 are met (Fix A docs, Fix B provisioner gate, Fix C preview-deploy comment, wrapper `--force-shared-parent` flag)
- [x] Feature works end-to-end as described — verified independently that the wrapper's new arg parser correctly handles `--force-shared-parent`, rejects unknown flags, rejects multiple positional args, and `--help` works
- [x] No scope creep — five precise changes (one per Fix + the wrapper plumbing + tests). No drive-by edits

### Code Quality
- [x] **Fix B placement** — gate sits at the right boundary: BEFORE the existing branch-lookup/secret-write logic, so a fail-fast exit short-circuits before any cross-contamination is possible
- [x] **Inline comment on the gate** is high-signal — names the exact failure mode (silent secret-mount cross-contam) AND the legitimate use cases for the escape hatch (paired collaboration, re-running cohorts)
- [x] **Error message is actionable**: identifies the branch + project, lists both likely causes, names the env var to set, and explicitly mentions the two legitimate cases. Workshop facilitators will know what to do
- [x] `rm -f "$create_response_file"` before `exit 1` — proper tempfile cleanup on the new error path; matches the existing cleanup-on-exit convention in this script
- [x] **Wrapper arg parser refactor** is the right shape: explicit `case "$1" in` loop preserves the strict "exactly one positional arg" property (the new `multiple positional arguments` error) AND adds `--*) ERROR: unknown flag` rejection
- [x] **Wrapper accepts both `--force-shared-parent` and `NEON_FORCE_SHARED_PARENT=true`** as env — via the `FORCE_SHARED_PARENT="${NEON_FORCE_SHARED_PARENT:-false}"` initialization. Either path works; the flag overrides the env. Documented
- [x] **Fix C preview-deploy.yml comment** correctly identifies that `'main'` is a literal-string fallback (verified against actual YAML — the `||` operator falls back to the literal string, not a "default branch" lookup); explains the consequence; cross-references the new docs section
- [x] **Fix A docs section** is well-scoped — branch namespace, types, multi-tenancy implications, what-to-expect-in-console, collision behavior. The "what you should expect to see" bullet generalizes the `va-worker`-confusion case from the downstream report into a verifiable invariant
- [x] markdownlint clean
- [x] actionlint clean (the new comment block in preview-deploy.yml passes)
- [x] shellcheck clean

### Testing
- [x] **Test 17 update** — correctly preserves the original idempotent-rerun assertion shape but updates it to require the `--force-shared-parent` flag, which is the realistic trigger for that path post-#90. Asserts the new log line that names `NEON_FORCE_SHARED_PARENT`, so a future regression that drops the flag-naming would fire
- [x] **Test 29 (new) — fail-fast path** — three assertions: error names branch+project, error names escape hatch, `gcloud.log` does NOT contain `secrets create database-url`. The third assertion is the load-bearing one — it's the actual cross-contam regression guard. Verified independently by tracing the script's gcloud calls under the fail-fast exit: the script exits 1 in Step 5.7b before reaching Step 5.7d's secrets calls
- [x] **Test 30 (new) — force-shared-parent path** — three assertions: log names the flag, no fail-fast error, downstream secret handling proceeds. Together with Test 29, the pair pins both directions of the gate
- [x] **All 7 test suites green**: 106 + 32 + 25 + 35 + 21 + 18 + 8 = 245 (was 239; +6 from the new tests + 1 carried-over from the Test 17 reshape)
- [x] CI green
- [ ] **One real test gap**: the wrapper's new arg-parser branches (`--unknown-flag`, multiple positional args, `--force-shared-parent` flag itself) are NOT exercised by `provision-workshop-roster.test.sh`. Manually verified all three; non-blocking but worth a follow-up

### Security
- [x] **No new credential surface** — the new env var is a boolean, not a secret
- [x] **The fail-fast path PREVENTS a security regression** (cross-contam: one attendee's `database-url` secret silently mounted to another's branch is a realistic data-confidentiality issue at scale). The cross-contam regression guard in Test 29 is the right framing
- [x] **The escape hatch is explicit** — never auto-set, never default-on. Inline comment in the script even calls this out: "Set explicitly; never default on"
- [x] **No injection vectors** — the new env var is read with `${NEON_FORCE_SHARED_PARENT:-false}` and compared with `!= "true"`. Standard shell-comparison pattern
- [x] **Tempfile cleanup on the new error path** preserves the script's no-leftover-tempfiles posture

### Recommendation
**GO**

Solid work on a real safety surface. The fail-or-force gate is the right shape: default behavior closes the cross-contam footgun, escape hatch preserves the legitimate use cases, error message is actionable. Test 29's `gcloud.log` cross-contam regression guard is the highest-signal assertion in the diff — it pins the property the gate is meant to enforce, not just the surface message.

Independently verified:
- Wrapper arg parser handles unknown flags, multiple positional args, and `--help` correctly (manual tests)
- Test 29's regression guard correctly traces through the script's flow under fail-fast (the exit happens before any `gcloud secrets` call)
- `'main'` fallback in `preview-deploy.yml` is a literal string, not a default-branch lookup — Fix C's comment is accurate
- All 7 test suites green; markdownlint + actionlint + shellcheck + pytest clean

None.

- **Wrapper arg-parser test gap.** The new branches in `provision-workshop-roster.sh` (unknown flag, multiple positional args, `--force-shared-parent` propagation) aren't exercised by `provision-workshop-roster.test.sh`. Three small additional tests would close the gap and make future arg-parser changes safer. Worth filing as a follow-up rather than expanding this PR's scope
- **The `--help` flag** is now a recognized branch in the wrapper. Worth adding it to the script's docstring header (currently only `--force-shared-parent` is documented as a flag) — pure docs polish
- **Forward-compat**: a future `provision-workshop-roster.sh` that adds another flag will need to update the `case` arm count + the `--help` `sed` range (currently `1,60p`). Worth noting in a header comment or a `# When adding flags, also bump the help range` comment
- (Carried) The original handoff doc referenced `va-worker` as an unexplained Neon branch in the user's console. Fix A's "what you should expect to see" bullet correctly generalizes this — but PLATFORM-GUIDE could use a one-line note: "Branch names match roster handles; if you see a branch named after a bot account, that's a roster row that used the bot's name as a handle." Pure docs; defer-able
- The escape-hatch flag name `--force-shared-parent` is descriptive but a little long. `--allow-existing-branch` or `--reuse-existing` might be more discoverable. Pure naming; current shape is fine

Issue #83, opened 2026-04-30
feat(scripts): add setup-solo-mode.sh as first-class bootstrap entry point


va-worker:
## Problem Statement

Multi-bot mode has `scripts/setup-accounts.sh` (374 lines, real provisioning that handles PAT capture, keyring storage, scope verification, account-flip recovery). **Solo mode has no equivalent.** A user who reads `CLAUDE.md` and runs `export AGILE_FLOW_SOLO_MODE=true` discovers — only after their first failed `/work-ticket` — that they also need to:

- Make sure `AGILE_FLOW_SOLO_MODE` is in shell rc (env vars don't survive across sessions)
- Restart Claude Code so the agent's snapshot env picks it up
- Unset `AGILE_FLOW_WORKER_ACCOUNT` and `AGILE_FLOW_REVIEWER_ACCOUNT` so the hook has no fallback
- Audit `GITHUB_PERSONAL_ACCESS_TOKEN*` env vars (which silently override `gh auth switch`)
- Confirm their personal account has `repo`, `project`, `workflow`, `read:project` scopes
- Run `gh auth refresh` if scopes are missing — and re-switch active account afterwards (gh refresh flips active account)
- Run `git config --local core.hooksPath scripts/hooks` (separate ticket #77 fixes this for the GCP path, but solo mode needs the same)
- Verify `gh api repos/<org>/<repo> --jq '.permissions.admin'` returns `true`

The downstream report at `tck517/tck517-app:docs/UPSTREAM-HANDOFF.md` Issue #4 documents seven distinct workarounds (A through G), every one of which would be eliminated by a single bootstrap script.

**Parent Epic:** vibeacademy/agile-flow-gcp#48
**Effort Estimate:** M
**Priority:** P0 — workshop-prep. Solo mode is the workshop-attendee path; without a bootstrap script, every attendee re-discovers the workarounds in the first 30 minutes.
**Surfaced:** 2026-04-30 dry-run on `tck517/tck517-app`.

---

## A. Environment Context

- New file: `scripts/setup-solo-mode.sh`
- Reference: existing `scripts/setup-accounts.sh` (multi-bot bootstrap) — match its style and quality bar
- Will be invoked by participants directly; should also be callable from `/bootstrap-workflow` (separate follow-up integration)
- `scripts/doctor.sh` should pass cleanly after `setup-solo-mode.sh` completes
[...]

Issue #82, opened 2026-04-30
feat(agents): make worker/reviewer agent briefs solo-mode aware; remove unconditional gh auth switch


va-worker:
## Problem Statement

`.claude/agents/github-ticket-worker.md:51` and `.claude/agents/pr-reviewer.md:74` both instruct the agent to run `gh auth switch --user {worker-bot}` (or `{reviewer-bot}`) as their first GitHub-touching action. This is wrong in two ways:

1. **Wrong in solo mode.** Solo-mode users have no bot accounts. The instruction either fails outright (no such account in the keyring) or — worse — silently switches to a stale leftover bot account, leading to opaque permission errors several operations later.
2. **Dangerous in any mode.** `gh auth switch` mutates global gh state visible to every other terminal the user has open. The downstream session at `tck517/tck517-app` flagged this as Workaround G: a worker subagent nearly ran the switch, which would have flipped the user's other shells mid-session.

`scripts/lint-agent-policies.sh:222` actively *requires* the string `gh auth switch` to be present in workflow agent files, so any fix here must ship together with a linter rule update (filed separately as #G).

Surfaced 2026-04-30 in the downstream `tck517/tck517-app:docs/UPSTREAM-HANDOFF.md` Issue #4. Solo-mode user reports: "This is the most consequential issue documented so far. Solo mode is the path most fork users will take."

**Parent Epic:** vibeacademy/agile-flow-gcp#48
**Effort Estimate:** S
**Priority:** P0 — every solo-mode user (workshop attendees, individual learners) hits this on the first agent invocation.
**Surfaced:** 2026-04-30 dry-run on `tck517/tck517-app`.

---

## A. Environment Context

- File: `.claude/agents/github-ticket-worker.md` lines 45-65 (the "CRITICAL: GitHub Account Identity" block + the bash example)
- File: `.claude/agents/pr-reviewer.md` lines 70-88 (the corresponding block)
- File: `scripts/lint-agent-policies.sh:222` — requires `gh auth switch\|GitHub Account Identity\|bot.*account` to be present
- Hook still does account switching as a PreToolUse hook for `gh pr create` and `gh pr review` — the agent doesn't need to do it itself

---

## B. Guardrails

- The replacement instruction must NOT call `gh auth switch` from agent context (mutates global state).
- Must work correctly in BOTH solo mode and multi-bot mode without conditionals in the prose. The simplest framing: "verify the active account, do not change it."
[...]

PR #101, opened 2026-04-30
fix(hook): make ensure-github-account.sh self-healing (#85)

### LINKED ISSUE #85: fix(hook): make ensure-github-account.sh self-healing — no-op when active account is non-bot
## Problem Statement

`.claude/hooks/ensure-github-account.sh` currently has a binary gate: if `AGILE_FLOW_SOLO_MODE=true`, no-op; else switch to `va-worker` / `va-reviewer`. This breaks down in three real scenarios:

1. **`AGILE_FLOW_SOLO_MODE` doesn't propagate into Claude Code subprocesses.** A user who exports the var AFTER starting their Claude Code session has the var visible in their interactive shell, but the agent's Bash subprocesses see a snapshot from session start. The hook reads `false`, switches to `va-worker`, and the user's "set one env var" expectation is silently violated.
2. **`gh auth refresh` flips the active account.** Solo-mode users who hit `INSUFFICIENT_SCOPES` and run `gh auth refresh` discover the refresh side-effects flip the active account to a bot. The hook then proceeds to "switch back" to a bot, undoing the user's manual recovery.
3. **Multi-bot mode does not have this problem in the documented flow, but multi-bot users who manually switch to their personal account for a one-off operation lose that switch on the next agent tool call.**

A safer logic is "if the user is already on a non-bot identity, respect that decision."

Surfaced 2026-04-30 in the downstream `tck517/tck517-app:docs/UPSTREAM-HANDOFF.md` Issue #4 (the active-account-flipping symptom + Workaround B). This is also the structural fix referenced as Fix B in the handoff.

**Effort Estimate:** S
**Priority:** P1 — defense-in-depth on top of #82 (agent briefs) and #83 (setup-solo-mode.sh). Useful even when both are merged; necessary if either is delayed.

---

## A. Environment Context

- File: `.claude/hooks/ensure-github-account.sh`
- File: `.claude/hooks/ensure-github-account.test.sh` — extend with the new no-op cases
- The hook reads `WORKER_ACCOUNT="${AGILE_FLOW_WORKER_ACCOUNT:-va-worker}"` and `REVIEWER_ACCOUNT="${AGILE_FLOW_REVIEWER_ACCOUNT:-va-reviewer}"`. The "non-bot account" check compares against this set.

---

## B. Guardrails

- **Backwards compatible.** Existing multi-bot setups (active account is `va-worker`, hook switches to `va-reviewer` for `gh pr review`) must continue to work unchanged.
- **No regression on solo mode.** `AGILE_FLOW_SOLO_MODE=true` continues to no-op as today (line 23 short-circuit stays).
- **Idempotent.** Running the hook on an already-correct active account is a no-op (no `gh auth status` mutation, no log line).
- **Safe failure mode.** If `gh auth status` fails or returns unparseable output, the hook should fail-safe (exit 0, no switch attempted) rather than fail-closed (exit 2, blocking the tool call). Solo-mode users with no gh auth at all should not be blocked from running non-gh commands.
- **Log when it would have switched but didn't.** When the hook detects "user on personal account; would switch but won't," emit a one-line stderr message so the user understands why the next operation may fail under the wrong identity. Quiet otherwise.

---

## C. Happy Path

Add a new short-circuit after the `AGILE_FLOW_SOLO_MODE` check and before the existing required-account logic:

```bash
# Self-healing: if the active account is NOT one of the configured bot
# accounts, respect that. Covers three cases:
#   1. AGILE_FLOW_SOLO_MODE=true was set after Claude Code started, so
#      the env var didn't propagate but the user IS on their personal
#      account.
#   2. `gh auth refresh` flipped active to a personal account; the
#      user's recovery should not be undone.
#   3. Multi-bot user manually switched for a one-off; they meant it.
ACTIVE=$(gh auth status --active --json login 2>/dev/null | jq -r '.login // empty')
if [[ -n "$ACTIVE" ]] && [[ "$ACTIVE" != "$WORKER_ACCOUNT" ]] && [[ "$ACTIVE" != "$REVIEWER_ACCOUNT" ]]; then
  echo "[ensure-github-account] Active account '$ACTIVE' is not a configured bot ($WORKER_ACCOUNT/$REVIEWER_ACCOUNT); not switching" >&2
  exit 0
fi
```

After this lands:
- Solo-mode-without-env-var users get the same effective behavior as solo mode
- `gh auth refresh` recoveries persist
- Multi-bot mode is unchanged (active is one of the bot accounts → existing logic runs)

---

## D. Definition of Done

- [ ] `.claude/hooks/ensure-github-account.sh` adds the self-healing block after the `AGILE_FLOW_SOLO_MODE` short-circuit
- [ ] `gh auth status --active` failure is handled (fail-safe to exit 0)
- [ ] `.claude/hooks/ensure-github-account.test.sh` adds 3+ new cases:
  - active account is `tck517` (a personal account), `AGILE_FLOW_SOLO_MODE` unset → no switch attempted
  - active account is `va-worker` (a configured bot), `gh pr review` triggered → switch to `va-reviewer` (existing flow)
  - `gh auth status` returns no active account → exit 0 silently
- [ ] All existing tests still pass
- [ ] shellcheck clean

---

## Why P1

This makes the hook more robust against real-world gh-state drift without changing the documented behavior. It's defense-in-depth on top of #82 and #83; if either of those are delayed, this alone meaningfully improves the solo-mode experience.

va-worker:
## Summary

Adds a keyring-state safety net to `.claude/hooks/ensure-github-account.sh`. The existing `AGILE_FLOW_SOLO_MODE` env-var check is the fast path; this is the defense-in-depth for when that env var didn't propagate into Claude Code's Bash tool subprocesses (the real-world failure mode that bit the 2026-04-30 dry-run on `tck517/tck517-app` — UPSTREAM-HANDOFF.md Issue #4 Fix B).

Two scenarios covered:
1. User ran `scripts/setup-solo-mode.sh` AFTER starting Claude Code. Env var is in their shell rc but Claude's Bash subprocesses see a session-start snapshot.
2. `gh auth refresh` flipped the active account to a personal account during scope recovery; switching back to a bot would silently undo the user's manual recovery.

## Implementation

- After fetching `current_account` (which the hook already does), check whether it matches one of the configured bots. If not, log to stderr and `exit 0`. Multi-bot mode is unchanged — `va-worker` active still switches to `va-reviewer` for `gh pr review`.
- Also added `|| true` to the existing `gh auth status | grep ... | sed ...` pipeline. Without it, `set -o pipefail` would kill the hook with exit 1 on unparsable `gh` state, blocking the tool call.
- Top-of-script docstring updated to describe both short-circuits (env-var fast path + keyring safety net).

## Deviation from ticket spec — small but worth noting

Ticket #85's Happy Path proposed `gh auth status --active --json login | jq -r '.login // empty'`. Empirically `gh auth status` does not support `--json login` — only `--jq` for filtering host-list JSON. I used the existing parsing pattern instead, which:

- Matches the test stub's output shape (no test fixture changes needed)
- Is consistent with the rest of the hook (line 59 already uses this pattern for the bot-switching path)
- Stays accurate to the ticket's intent — same self-heal semantics, different mechanic

## Tests

- **Test 5:** personal account active + `AGILE_FLOW_SOLO_MODE` unset → no `gh auth switch` (the originating dry-run failure mode)
- **Test 6:** personal account active + `AGILE_FLOW_SOLO_MODE=false` → still no switch (defends against shells that set the var to non-`true` values)
- **Test 7:** `va-worker` active + `gh pr review` → switches to `va-reviewer` (multi-bot regression guard)
- **Test 8:** empty active account → self-heal does NOT short-circuit; falls through to existing logic (fail-open)

8 tests / 17 assertions — all pass.

## Test plan

- [x] `./.claude/hooks/ensure-github-account.test.sh` — 17/17
- [x] `shellcheck` clean (with project's standard exclusions)
- [x] All validation scripts pass
- [x] `pytest` 22/22
- [ ] CI green on PR
- [ ] Real-world: next mid-session `setup-solo-mode.sh` run on a fresh fork (the upcoming dry-run will exercise this path)

Closes #85

### DIFF
--- a/.claude/hooks/ensure-github-account.sh
+++ b/.claude/hooks/ensure-github-account.sh
@@ -12,6 +12,16 @@
 #
 # Configure via environment variables or edit the defaults below.
 #
+# Two short-circuits before the hook attempts any account switch:
+#   1. Env-var fast path — AGILE_FLOW_SOLO_MODE=true exits 0 unconditionally.
+#   2. Keyring safety net (#85) — if the active gh account is not one of the
+#      configured bots, exit 0. Catches solo-mode setups where the env var
+#      didn't propagate to Claude's Bash subprocesses, and post
+#      `gh auth refresh` recoveries that flipped the active account to
+#      personal. Both are present because env vars and keyring state can
+#      diverge in real-world scenarios; the keyring is the reliable signal
+#      when they conflict. See UPSTREAM-HANDOFF.md Issue #4 Fix B.
+#
 
 set -euo pipefail
 
@@ -55,8 +65,41 @@ case "$tool_name" in
     ;;
 esac
 
-# Get current active account
-current_account=$(gh auth status 2>&1 | grep -B2 "Active account: true" | grep "Logged in to" | sed 's/.*account \([^ ]*\).*/\1/')
+# Get current active account. Fail-open: with `set -o pipefail`, an
+# empty/unparsable `gh auth status` would otherwise kill the hook
+# with exit 1 and block the tool call. The self-heal block below and
+# the existing switch logic both handle empty `current_account`
+# gracefully, so we'd rather get an empty value than die here.
+current_account=$(gh auth status 2>&1 | grep -B2 "Active account: true" | grep "Logged in to" | sed 's/.*account \([^ ]*\).*/\1/' || true)
+
+# Self-healing safety net (#85). If the active account is not one of
+# the configured bots, the user is on a personal account on purpose
+# (solo mode without env-var propagation, or post `gh auth refresh`
+# recovery). Respect that and exit 0 — don't undo their state.
+#
+# Two real-world scenarios where this fires:
+#   1. User ran scripts/setup-solo-mode.sh AFTER starting Claude Code.
+#      AGILE_FLOW_SOLO_MODE is set in their shell rc, but Claude Code's
+#      Bash subprocesses see a snapshot from session start. The
+#      env-var fast path at the top of this hook misses; the keyring
+#      is the more reliable signal.
+#   2. `gh auth refresh` flipped the active account to a personal
+#      account during scope recovery. Switching back to a bot here
+#      would silently undo the user's manual recovery.
+#
+# Multi-bot mode is unaffected: a user with $WORKER_ACCOUNT active
+# falls through to the existing switch-to-reviewer logic below.
+# Fail-open if `gh auth status` returned no parsable account — let
+# the existing logic surface the gh-state error rather than blocking
+# the tool call here.
+# See: docs/UPSTREAM-HANDOFF.md Issue #4 Fix B (originating fork's
+# 2026-04-30 dry-run).
+if [[ -n "$current_account" ]] \
+  && [[ "$current_account" != "$WORKER_ACCOUNT" ]] \
+  && [[ "$current_account" != "$REVIEWER_ACCOUNT" ]]; then
+  echo "[ensure-github-account] Active account '$current_account' is not a configured bot ($WORKER_ACCOUNT/$REVIEWER_ACCOUNT); not switching for $tool_name" >&2
+  exit 0
+fi
 
 # Switch if needed
 if [[ "$current_account" != "$required_account" ]]; then
--- a/.claude/hooks/ensure-github-account.test.sh
+++ b/.claude/hooks/ensure-github-account.test.sh
@@ -174,6 +174,157 @@ else
   FAIL=$((FAIL + 1))
 fi
 
+# ── Test 5: Self-heal — personal account active, env var unset ─────────
+# The downstream's #85 failure mode: user runs setup-solo-mode.sh
+# mid-session, env var doesn't propagate to subprocesses, but the
+# keyring shows them on a personal account. Hook must respect that
+# and not flip them to a bot.
+
+echo ""
+echo "Test 5: personal account active + AGILE_FLOW_SOLO_MODE unset → no switch"
+
+T5=$(new_tmp)
+make_gh_stub "$T5" "tck517"
+
+input='{"tool_name":"Bash","tool_input":{"command":"gh pr create --title foo"}}'
+
+set +e
+PATH="$T5/bin:$PATH" \
+  AGILE_FLOW_WORKER_ACCOUNT="va-worker" \
+  AGILE_FLOW_REVIEWER_ACCOUNT="va-reviewer" \
+  bash "$HOOK" <<< "$input" > "$T5/stdout.log" 2>&1
+ec=$?
+set -e
+
+assert_eq "0" "$ec" "exit 0 (self-heal exits cleanly)"
+
+if [[ -f "$T5/gh.log" ]] && grep -q "auth switch" "$T5/gh.log"; then
+  echo -e "  ${RED}✗${NC} gh auth switch was called — self-heal failed to short-circuit"
+  FAIL=$((FAIL + 1))
+else
+  echo -e "  ${GREEN}✓${NC} no gh auth switch attempted (self-heal respected personal account)"
+  PASS=$((PASS + 1))
+fi
+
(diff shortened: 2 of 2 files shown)


va-worker:
## PR Review — #101

### Requirements

DoD walkthrough against #85:
- Self-healing block added (lines 75–102 of `ensure-github-account.sh`) — placed after the existing `current_account` lookup rather than the ticket's "after solo-mode / before required-account" wording, but semantically equivalent. The placement is actually a small win: it reuses the same parse the bot-switching path already does, so no second `gh auth status` call.
- `gh auth status` failure handled via `|| true` on the existing parse pipeline (line 73). Solo-mode users with broken / no `gh` auth no longer get blocked from non-`gh pr` Bash calls.
- Three+ new test cases added (Tests 5, 6, 7, 8 — four cases, exceeds the ticket's "3+").
- All existing tests still pass (Tests 1–4 unchanged).
- `shellcheck` clean — verified locally on both files (exit 0).
- Stderr log line present: `[ensure-github-account] Active account 'X' is not a configured bot (Y/Z); not switching for <tool_name>`.

### Code Quality

The deviation from the ticket's `gh auth status --active --json login | jq` Happy Path is justified. I empirically reproduced the failure: `gh auth status --active --json login` returns `Unknown JSON field: "login"` (exit 1) — `gh auth status` only supports `hosts` as a JSON field. The PR's substitution (the existing `grep -B2 ... | sed ...` pipeline) matches the test stub's output shape, is consistent with line 73 of the hook itself, and preserves the ticket's intent (self-heal semantics) rather than its mechanic. Documented in the PR description; appropriate.

The `|| true` addition is correctly scoped — minimal change, fixes a real fail-closed bug exposed by Test 8, and the alternative (removing `set -o pipefail`) would have been more invasive. Inline comment explains the reasoning.

### Testing
- [x] All tests pass in CI

Verified locally: `./.claude/hooks/ensure-github-account.test.sh` → 17/17 pass. CI: 10 jobs green, 2 deployment jobs `skipping` as expected for hook-only changes.

Test quality is genuinely good. Test 7 is the regression guard the deviation makes essential — it stubs `va-worker` active, sends a `gh pr review` command, and asserts `gh auth switch --user va-reviewer` appears in the gh-call log. That actually proves the multi-bot path is unaffected, not just that exit code is 0. Test 8 doubles as documentation of the deliberate fail-open choice for empty parse output.

### Security

`$current_account` is interpolated into a stderr echo and `[[ ]]` string comparisons — never `exec`'d or passed to a subshell. Safe.

### Recommendation
**GO**

Clean, focused fix for a real-world failure mode (UPSTREAM-HANDOFF.md Issue #4 Fix B). The deviation from the ticket's exact implementation is well-reasoned and empirically grounded. Test coverage is meaningful; CI is green; shellcheck is clean. The bonus `|| true` fix-along is appropriately scoped and well-commented. Multi-bot mode regression is explicitly guarded by Test 7.

None.

- **Test 8 vs ticket guardrail wording.** The ticket's "Safe failure mode" guardrail says "If `gh auth status` fails or returns unparseable output, the hook should fail-safe (exit 0, no switch attempted)." Test 8 documents the opposite choice — let the existing logic attempt the switch and surface the gh-state error. The PR author's reasoning (surface real auth errors rather than mask them) is defensible, and the test passes only because the stub's `auth switch` returns success; in a real environment with no auth at all, this path would exit 2. Worth a follow-up discussion on whether the ticket's strict fail-safe is the right behavior, or whether the PR's "fail-open here, let the switch attempt fail later" is. Either way, not a blocker for #85's primary objective.
- **Stderr message includes `$tool_name` (good)** — small touch that helps users understand which call was inspected. No change needed; calling out as a quality nit.
- Consider, in a future ticket, gating the env-var fast path on the same keyring check (or removing the env-var fast path entirely once #82 + #83 land in a stable release) so there's a single source of truth for "is this solo mode?". The current two-short-circuit design is correct for backwards compatibility but adds two places that need to agree.

PR #97, opened 2026-04-30
feat(devcontainer): add GitHub Codespace as recommended workshop setup (closes #95)

va-worker:
## Summary

Closes #95. Adds a GitHub Codespace devcontainer as the recommended workshop setup path. Eliminates Windows-shell incompatibility, tool-version drift across attendees, and most of the solo-mode env-var-propagation friction at the environment level.

## What changed

| File | Change |
|---|---|
| `.devcontainer/devcontainer.json` (new) | Ubuntu 24.04 LTS base + features (Python 3.12, Node 20, gh, gcloud, uv); `AGILE_FLOW_SOLO_MODE=true` in `containerEnv` (propagates to all agent subprocesses); `postCreateCommand` runs `setup-solo-mode.sh` and `uv sync --extra dev`; 2-core/4gb default machine; port 8080 forwarded |
| `README.md` | Open in Codespaces badge + one-line "New here?" callout |
| `docs/GETTING-STARTED.md` Step 1 | Split into Path A (Codespace, recommended) and Path B (local clone, fallback). Closed-network attendees keep the manual path |
| `docs/PLATFORM-GUIDE.md` | New "Codespace cost estimate" subsection inside Workshop Lifecycle: per-attendee compute cost (~$15-25 per cohort), free-tier coverage, hibernate-on-idle behavior, closed-network fallback pointer |

## Why

Workshop attendees today follow `docs/GETTING-STARTED.md` to set up their machines. The current path requires platform-correct installs of `gh`, `gcloud`, `python3.12`, `uv`, `node`, plus shell-rc env vars, plus `gh auth login`, plus the per-clone `git config core.hooksPath` step (#77), plus `setup-solo-mode.sh` (#83). The 2026-04-29 facilitator session lost time to Windows-side path/shell incompatibilities.

A Codespace eliminates the whole surface: every attendee gets the same Linux, same tools, pre-authenticated gh, `AGILE_FLOW_SOLO_MODE=true` in `containerEnv` so it propagates into agent subprocesses without a Claude Code restart, and `setup-solo-mode.sh` runs automatically as `postCreateCommand`.

## Trade-offs

| Concern | Decision |
|---|---|
| Cost (~$15-25 per cohort compute) | Documented in PLATFORM-GUIDE; mostly within per-attendee free tier |
| Closed-network attendees | Path B (local clone) remains first-class; framework supports both |
| Third-party gcloud feature (`dhoeric/features`) | Acceptable for workshop scope; no first-party gcloud feature exists |
[...]

PR #110, opened 2026-05-01
fix(setup): solo-mode bootstrap survives Codespaces GITHUB_TOKEN (#102)

### LINKED ISSUE #102: fix(setup): solo-mode bootstrap fails in Codespaces because gh refuses to refresh GITHUB_TOKEN
## Symptom

In a fresh GitHub Codespace launched from a fork of this template, the
devcontainer's `postCreateCommand` (`bash scripts/setup-solo-mode.sh && uv sync --extra dev`)
fails at Step 4/8 with:

```
! Missing scopes: repo project workflow read:project
→ Running: gh auth refresh -h github.com -s repo project workflow read:project

The value of the GITHUB_TOKEN environment variable is being used for authentication.
To refresh credentials stored in GitHub CLI, first clear the value from the environment.
✗ gh auth refresh failed
postCreateCommand from devcontainer.json failed with exit code 1.
Skipping any further user-provided commands.
```

The Codespace finishes "configuring" but is unusable — `uv sync` never
ran, the pre-push hook was never activated, and there is no clear path
forward documented for the user.

This is the **first failure mode after a user follows the framework's
recommended workshop setup** (PR #97 — devcontainer + Codespace
positioning).

## Root cause

Codespaces auto-provisions a `GITHUB_TOKEN` env var into every
container — that's how the pre-installed `gh` CLI works without a
manual login. The token is a **GitHub user-to-server token** (`ghu_...`
prefix) tied to the Codespace's lifetime and **refresh-impossible by
design**: `gh auth refresh` requires write access to a token gh owns,
and gh does not own this one.

`scripts/setup-solo-mode.sh` Step 4/8 detects missing scopes correctly
and calls `gh auth refresh` to add them. The refresh fails. The script
exits 1. Codespaces aborts the rest of the post-create chain — not
just the refresh step, but `uv sync --extra dev` too.

This is the **same class of bug** as the original Issue #4 (token-
precedence problem from `GITHUB_PERSONAL_ACCESS_TOKEN*` shadowing the
keyring), wearing different clothes. Step 3/8's audit specifically
checks for `GITHUB_PERSONAL_ACCESS_TOKEN*` — Codespaces is clean of
those, so Step 3 passes — but the *category* of "env-var token that
overrides keyring" is broader than that one variable name. `GITHUB_TOKEN`
is the Codespaces-specific instance.

## Evidence

Direct quote from `/workspaces/.codespaces/.persistedshare/creation.log`:

```
2026-05-01 11:20:46.166Z: → Added AGILE_FLOW_SOLO_MODE to /home/vscode/.bashrc

--- Step 3/8: Audit GITHUB_PERSONAL_ACCESS_TOKEN env vars ---
2026-05-01 11:20:46.203Z: ✓ No token env vars detected (gh keyring is the source of truth)

--- Step 4/8: Verify gh token scopes ---
2026-05-01 11:20:46.606Z: ! Missing scopes: repo project workflow read:project
→ Running: gh auth refresh -h github.com -s repo project workflow read:project

2026-05-01 11:20:51.702Z: The value of the GITHUB_TOKEN environment variable is being used for authentication.
To refresh credentials stored in GitHub CLI, first clear the value from the environment.
2026-05-01 11:20:51.713Z: ✗ gh auth refresh failed
```

`gh auth status` from inside the resulting (broken) Codespace:

```
github.com
  ✓ Logged in to github.com account tck517 (GITHUB_TOKEN)
  - Active account: true
  - Token: ghu_************************************
```

Note the `(GITHUB_TOKEN)` annotation — gh explicitly tells the user
the active credential is from the env var, not the keyring. The token
prefix `ghu_` confirms it's a user-to-server token.

`env | grep -c GITHUB_PERSONAL_ACCESS_TOKEN` returns `0` — Step 3's
audit is correct. The fix needs to extend the audit, not replace it.

## Proposed fix

Three complementary changes; ship them together.

### Fix A — Detect Codespaces in `setup-solo-mode.sh` Step 4 and route around `gh auth refresh`

The `CODESPACES=true` env var is set by Codespaces automatically.
Use it to gate the refresh attempt:

```bash
# In Step 4/8: Verify gh token scopes
if [[ -n "${CODESPACES:-}" ]]; then
  print_warning "Codespaces detected — GITHUB_TOKEN is read-only and cannot be refreshed."
  print_info "To get full scopes (repo, project, read:project, workflow):"
  echo ""
  echo "  1. Create a Codespaces user secret named GH_TOKEN at:"
  echo "     https://github.com/settings/codespaces"
  echo "  2. Generate a fine-grained PAT with repo + project access at:"
  echo "     https://github.com/settings/tokens?type=beta"
  echo "  3. Grant the secret access to your fork."
  echo "  4. Rebuild the Codespace (Command Palette → Codespaces: Rebuild Container)."
  echo ""
  print_info "GH_TOKEN takes precedence over GITHUB_TOKEN, so gh will use the"
  print_info "broader-scoped token after rebuild without further changes."
  echo ""
  # Don't fail the script — the user has been told what to do, and
  # uv sync should still run so the rest of the env is usable.
  exit 0
fi

# (existing scope-refresh logic for non-Codespaces shells continues below)
```

### Fix B — Generalize Step 3's audit to *any* token env var that gh recognizes

`GITHUB_PERSONAL_ACCESS_TOKEN*` is one example of "env-var token
overriding the keyring." There are at least three others gh respects:
`GITHUB_TOKEN`, `GH_TOKEN`, `GH_ENTERPRISE_TOKEN`. The audit should
flag all of them, with severity tiered:

```bash
# Step 3 — generalized audit
declare -a found_tokens
for var in GITHUB_TOKEN GH_TOKEN GH_ENTERPRISE_TOKEN \
           GITHUB_PERSONAL_ACCESS_TOKEN \
           $(env | grep -oE '^GITHUB_PERSONAL_ACCESS_TOKEN[A-Z_]*' | sort -u); do
  if [[ -n "${!var:-}" ]]; then
    found_tokens+=("$var")
  fi
done

if [[ ${#found_tokens[@]} -eq 0 ]]; then
  print_success "No token env vars detected (gh keyring is the source of truth)"
elif [[ -n "${CODESPACES:-}" ]] && [[ "${found_tokens[*]}" == "GITHUB_TOKEN" ]]; then
  # Expected case — single GITHUB_TOKEN in Codespaces. Informational, not a fail.
  print_info "Codespaces GITHUB_TOKEN detected (expected). See Step 4 for scope guidance."
else
  print_warning "Token env vars detected: ${found_tokens[*]}"
  print_warning "These override gh's keyring and may shadow account switches."
  print_info "Recommended: remove from your shell rc and rotate."
fi
```

### Fix C — Make the post-create command tolerant

Even with Fix A, `uv sync` should run regardless of whether the
solo-mode setup completed cleanly. The current chain
(`bash scripts/setup-solo-mode.sh && uv sync --extra dev`)
short-circuits on the first failure. Recommend:

```json
"postCreateCommand": "bash scripts/setup-solo-mode.sh; uv sync --extra dev"
```

(`;` instead of `&&`.) Trade-off: a hard error in setup-solo-mode is
no longer fatal to container provisioning. But: setup-solo-mode is
*advisory* — it persists an env var and audits state. uv sync is
*essential* — without it, the Python deps aren't installed and nothing
in the repo runs. Don't gate the latter on the former.

## Verification

Three scenarios to confirm fixed:

1. **Fresh Codespace, no `GH_TOKEN` secret configured:** post-create
   succeeds. `setup-solo-mode.sh` exits 0 with the user-actionable
   "configure GH_TOKEN secret" message. `uv sync` runs. The user can
   open a terminal and start working immediately.

2. **Fresh Codespace WITH `GH_TOKEN` secret configured (workshop happy path):**
   `gh auth status` shows `GH_TOKEN` (not `GITHUB_TOKEN`) as the active
   credential, with all required scopes. Step 4 passes silently. The
   rest of the bootstrap proceeds as designed.

3. **Local terminal (no Codespaces):** `CODESPACES` env var is unset.
   Step 4 falls through to the existing `gh auth refresh` logic. No
   regression to non-Codespaces flows.

## Notes for upstream session

- The "workshop happy path" the framework recommends is now Codespace-
  based per PR #97. **This bug breaks that path on the first run.** Every
  workshop attendee will hit this until fixed. Priority should reflect
  that.

- The fix isn't novel territory — `setup-solo-mode.sh` already has
  Codespaces-conditional comments in the script header (PR #97 added
  them); the script just doesn't actually branch on `CODESPACES` at
  Step 4. The infrastructure is already there; what's missing is a
  guard.

- Optional but recommended: update `docs/GETTING-STARTED.md`'s
  Codespaces section to walk through the GH_TOKEN secret setup as the
  documented happy path. Right now PR #97 mentions devcontainer support
  but doesn't address the token-scope gap in any user-facing doc.

- Adjacent observation: the Codespaces-provisioned `GITHUB_TOKEN`
  doesn't include `project` scope by default, which means even AFTER
  the Codespace is unblocked, `/bootstrap-workflow` cannot create a
  project board against it. The GH_TOKEN-secret workaround is the
  ONLY supported path for Codespaces users who want the full agent
  workflow. Worth being explicit about.

---

Originating fork: `vibeacademy/tck517-app-g` (Codespace).
Discovered: 2026-05-01 during the second dry-run of the patched template
(first dry-run filed UPSTREAM-HANDOFF.md Issues #1–#5; Issue #4 Fix B
was filed separately as #85 and is also fixed).

va-worker:
## Summary

Three complementary changes for the failure mode that bricked Codespace provisioning on the 2026-05-01 dry-run on `tck517/tck517-app-g`. Closes #102.

**Fix A** — `scripts/setup-solo-mode.sh` Step 4 now detects `CODESPACES` env var directly. Codespaces' auto-injected `GITHUB_TOKEN` is a user-to-server token (`ghu_` prefix) that `gh` does NOT own; `gh auth refresh` fails against it regardless of TTY state. The earlier #97 `IS_INTERACTIVE` guard inferred from TTY presence — empirically Codespaces postCreateCommand DOES have a TTY, so the guard didn't fire. New CODESPACES branch prints attendee-actionable guidance (configure `GH_TOKEN` Codespaces user secret, fine-grained PAT, rebuild) and continues cleanly.

**Fix B** — `scripts/setup-solo-mode.sh` Step 3's audit generalized to cover `GITHUB_TOKEN` and `GH_ENTERPRISE_TOKEN` (in addition to existing `GITHUB_PERSONAL_ACCESS_TOKEN*` and `GH_TOKEN`). All four belong to the "env-var token that shadows the keyring" class. Special case: in Codespaces, a single `GITHUB_TOKEN` is expected — downgrade to informational rather than scare-warning. Outside Codespaces, all four still warn.

**Fix C** — `.devcontainer/devcontainer.json` `postCreateCommand` changed from `&&` to `;` between `setup-solo-mode.sh` and `uv sync --extra dev`. The two are independent: setup-solo-mode is advisory; uv sync is essential.

## Why each fix

- Without Fix A: every workshop attendee opening a Codespace hits `✗ gh auth refresh failed` and the postCreateCommand exits 1.
- Without Fix B: Codespaces' `GITHUB_TOKEN` was already shadowing the keyring on every workshop fork; the audit just wasn't checking for it.
- Without Fix C: even with Fix A in place, any non-fatal warning from setup-solo-mode would block uv sync, leaving the Python environment unprovisioned.

## Tests

42/42 pass locally (28 prior + 14 new):

- **Test 9** — Codespaces + missing scopes → WARN+continue, no `gh auth refresh` attempted, attendee gets specific GH_TOKEN guidance
- **Test 10** — Codespaces + only GITHUB_TOKEN set → Step 3 informational, not scary warning
- **Test 11** — Non-Codespace + GITHUB_TOKEN in env → Step 3 still warns (special case is properly narrow)

## Test plan

- [x] `./scripts/setup-solo-mode.test.sh` — 42/42
- [x] shellcheck on script + tests
- [x] devcontainer.json parses (after stripping JSONC comments)
- [x] All other validators (json/version/commands/agents/policies/lint-agent-policies/agent-restrictions)
- [x] actionlint clean
- [x] pytest 22/22
- [ ] CI green on PR
- [ ] Real-world: next Codespace boot on a workshop fork should provision cleanly through `setup-solo-mode.sh` → `uv sync` even without a `GH_TOKEN` secret configured

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -59,7 +59,12 @@
   // GITHUB_TOKEN, so gh is already authenticated; setup-solo-mode.sh
   // does the rest (hooks path, scope check, admin verification). uv sync
   // installs Python deps so `uv run pytest` works on first command.
-  "postCreateCommand": "bash scripts/setup-solo-mode.sh && uv sync --extra dev",
+  //
+  // Use `;` (not `&&`) so a setup-solo-mode warning or non-fatal exit
+  // doesn't block uv sync. setup-solo-mode is advisory (audits state,
+  // persists an env var); uv sync is essential (without Python deps
+  // nothing runs). Don't gate the latter on the former. (#102 Fix C.)
+  "postCreateCommand": "bash scripts/setup-solo-mode.sh; uv sync --extra dev",
 
   // FastAPI dev server.
   "forwardPorts": [8080],
--- a/scripts/setup-solo-mode.sh
+++ b/scripts/setup-solo-mode.sh
@@ -217,22 +217,42 @@ echo ""
 # In solo mode, ANY of these env vars override `gh auth switch` silently.
 # Detection is from the current shell environment (what gh sees right now)
 # AND from the shell rc file (so the user knows what to remove permanently).
+#
+# Tracked vars (#102 generalization): the broader category is "env-var
+# token that shadows the keyring." `gh` recognizes GITHUB_TOKEN, GH_TOKEN,
+# GH_ENTERPRISE_TOKEN, and the GITHUB_PERSONAL_ACCESS_TOKEN* family.
+# All four belong to the same class.
 token_env_vars=()
 while IFS= read -r line; do
     [ -n "$line" ] && token_env_vars+=("$line")
-done < <(env | grep -E "^(GITHUB_PERSONAL_ACCESS_TOKEN|GH_TOKEN)" | cut -d= -f1 || true)
+done < <(env | grep -E "^(GITHUB_PERSONAL_ACCESS_TOKEN|GH_TOKEN|GITHUB_TOKEN|GH_ENTERPRISE_TOKEN)" | cut -d= -f1 || true)
 
 token_rc_lines=()
 if [ -f "$profile" ]; then
     while IFS= read -r line; do
         [ -n "$line" ] && token_rc_lines+=("$line")
-    done < <(grep -nE "^(export\s+)?GITHUB_PERSONAL_ACCESS_TOKEN|^(export\s+)?GH_TOKEN|^set -Ux\s+GITHUB_PERSONAL_ACCESS_TOKEN|^set -Ux\s+GH_TOKEN" "$profile" 2>/dev/null || true)
+    done < <(grep -nE "^(export\s+)?(GITHUB_PERSONAL_ACCESS_TOKEN|GH_TOKEN|GITHUB_TOKEN|GH_ENTERPRISE_TOKEN)|^set -Ux\s+(GITHUB_PERSONAL_ACCESS_TOKEN|GH_TOKEN|GITHUB_TOKEN|GH_ENTERPRISE_TOKEN)" "$profile" 2>/dev/null || true)
 fi
 
 env_count=${#token_env_vars[@]}
 rc_count=${#token_rc_lines[@]}
+
+# Codespaces auto-injects GITHUB_TOKEN. If that's the ONLY token var
+# present, treat it as expected/informational rather than a warning —
+# noise is the enemy of attendees finding the real signal. Step 4
+# handles the scope-refresh workaround for the Codespaces case.
+expected_codespace_only=false
+if [ -n "${CODESPACES:-}" ] \
+   && [ "$env_count" -eq 1 ] \
+   && [ "${token_env_vars[0]}" = "GITHUB_TOKEN" ] \
+   && [ "$rc_count" -eq 0 ]; then
+    expected_codespace_only=true
+fi
+
 if [ "$env_count" -eq 0 ] && [ "$rc_count" -eq 0 ]; then
     print_success "No token env vars detected (gh keyring is the source of truth)"
+elif [ "$expected_codespace_only" = "true" ]; then
+    print_info "Codespaces GITHUB_TOKEN detected (expected). Step 4 will check its scopes."
 else
     print_warning "Found token env vars that override gh auth switch:"
     echo ""
@@ -278,11 +298,41 @@ done
 
 if [ ${#missing_scopes[@]} -eq 0 ]; then
     print_success "All required scopes present (${required_scopes[*]})"
+elif [ -n "${CODESPACES:-}" ]; then
+    # Codespaces (#102): the auto-injected GITHUB_TOKEN is a user-to-server
+    # token (`ghu_` prefix) that gh does NOT own. `gh auth refresh` against
+    # it fails with "The value of the GITHUB_TOKEN environment variable is
+    # being used for authentication. To refresh credentials stored in
+    # GitHub CLI, first clear the value from the environment." This is true
+    # regardless of TTY presence (#97's IS_INTERACTIVE guard doesn't help
+    # here because postCreateCommand may have a TTY but the refresh still
+    # fails on the same error).
+    #
+    # The supported workshop path is to configure a GH_TOKEN Codespaces
+    # user secret with a fine-grained PAT carrying the required scopes.
+    # gh prefers GH_TOKEN over GITHUB_TOKEN, so once that secret is set
+    # and the Codespace is rebuilt, scopes are satisfied without any
+    # refresh call.
+    print_warning "Missing scopes: ${missing_scopes[*]}"
+    print_warning "Codespaces detected — GITHUB_TOKEN cannot be refreshed (gh does not own it)."
+    echo ""
+    print_info "To fix: configure a GH_TOKEN Codespaces user secret with the required scopes."
+    print_info "  1. Generate a fine-grained PAT at https://github.com/settings/tokens?type=beta"
+    print_info "     Required scopes: ${required_scopes[*]}"
+    print_info "  2. Save as a Codespaces user secret named GH_TOKEN at"
+    print_info "     https://github.com/settings/codespaces"
+    print_info "  3. Grant the secret access to this fork"
+    print_info "  4. Rebuild the Codespace (Command Palette → Codespaces: Rebuild Container)"
+    echo ""
+    print_info "GH_TOKEN takes precedence over GITHUB_TOKEN, so gh will use the"
+    print_info "broader-scoped token after rebuild without further changes."
+    echo ""
+    print_warning "Continuing — uv sync and the rest of postCreate should still run."
 elif [ "$IS_INTERACTIVE" != "true" ]; then
-    # Non-interactive (Codespace postCreateCommand, CI, piped invocation):
-    # `gh auth refresh` would hang waiting for browser OAuth. WARN and
-    # surface the manual command for the user to run from their first
-    # interactive terminal. See #97.
+    # Non-interactive non-Codespace (CI, piped invocation): `gh auth
+    # refresh` would hang waiting for browser OAuth. WARN and surface the
+    # manual command for the user to run from their first interactive
+    # terminal. See #97.
     print_warning "Missing scopes: ${missing_scopes[*]}"
     print_warning "Skipping refresh (non-interactive context — would block on browser OAuth)"
     print_info "Run manually after this script finishes:"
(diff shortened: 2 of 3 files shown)


va-worker:
## PR Review — #110

### Requirements

All three fixes from #102 ship together as proposed. Fix A (CODESPACES branch in Step 4) is positioned correctly between the "all scopes present" and "non-interactive" branches; Fix B (Step 3 audit generalization + narrow Codespaces special case); Fix C (`&&` → `;` in `postCreateCommand`).

### Code Quality

The CODESPACES branch matches the shape of the existing `IS_INTERACTIVE` branch (warn + continue, surface a manual fix command). The `expected_codespace_only` flag is a clean way to express the four-way conjunction. Both URLs in the print_info messages are GitHub-canonical (`https://github.com/settings/tokens?type=beta`, `https://github.com/settings/codespaces`). Inline comments at lines 220–224, 240–243, and 302–315 explain the *why* — token-class generalization, noise-reduction motivation, and the empirical TTY-doesn't-help finding respectively. Code reads as deliberate, not over-engineered.

### Testing
- [x] All tests pass in CI

Verified locally: **42/42 pass (28 prior + 14 new)** as the author claims. Tests 9–11 are well-targeted:

- **Test 9** asserts the CODESPACES branch exits 0, names the context, names the workaround (GH_TOKEN secret + fine-grained PAT + rebuild), AND critically that `gh auth refresh` was *not* invoked (greps `$T9/gh.log` for absence). This last assertion is what guarantees the Codespace doesn't get bricked.
- **Test 10** locks in the Step 3 informational-vs-warning distinction for the Codespaces happy-token-state case.
- **Test 11** is the necessary mirror — outside Codespaces, GITHUB_TOKEN in env still warns. Without this test the special case could silently drift permissive.

### Security

Token values from env vars are never echoed; only var *names* are surfaced. The only new IO is print_info with static GitHub URLs.

### Verifications performed

1. **TTY claim** — agreed. The author's empirical finding that Codespaces postCreateCommand has a TTY (so #97's `IS_INTERACTIVE` guard didn't fire) matches the evidence in #102: the failure log shows `→ Running: gh auth refresh ...` which is from the interactive branch. Detecting `CODESPACES` directly is the correct fix.
2. **Idempotency on re-run** — when GH_TOKEN secret is correctly configured, gh prefers GH_TOKEN, so its scopes show in `gh auth status`. If complete: `missing_scopes` is empty, the first branch fires (`All required scopes present`). If incomplete: the CODESPACES branch fires again with the same actionable guidance. Logic is correct.
3. **`expected_codespace_only` narrowness** — requires `CODESPACES` set AND `env_count == 1` AND the sole var is exactly `GITHUB_TOKEN` AND `rc_count == 0`. A user who has `GITHUB_TOKEN` in `.bashrc` will trip `rc_count > 0`, the special case won't apply, and the standard warning surfaces. Test 11 anchors the non-Codespaces leg.
4. **Step 4 branch order** — `all scopes present` → `CODESPACES` → `IS_INTERACTIVE != true` → interactive `else`. Correct.
5. **Devcontainer change** — only the `&&` → `;` swap; the new comment explains why. JSONC parses.
6. **shellcheck** — clean on both scripts.
7. **CI** — all 12 jobs green.

### Recommendation
**GO**

This is a tight, well-scoped fix for a P0 workshop blocker. The diagnosis in #102 is unusually thorough (token type, `ghu_` prefix, evidence from creation.log, env audit), the three fixes are independently motivated and complementary, and the test coverage anchors each behavioral change including the failure-mode where it matters most (`gh auth refresh` MUST NOT be called in Codespaces). Ship it before the next dry-run.

None.

- **Doc follow-up** — #102's "Notes for upstream session" suggests updating `docs/GETTING-STARTED.md`'s Codespaces section to walk through GH_TOKEN secret setup as the documented happy path. Worth a separate ticket; not blocking on this PR. The script-level guidance is already attendee-actionable.
- **Adjacent observation from #102** — even after this fix, Codespaces-provisioned `GITHUB_TOKEN` lacks `project` scope, so `/bootstrap-workflow` cannot create a project board against it. The GH_TOKEN-secret path is the only route to a fully-working agent flow in Codespaces. Worth being explicit about in workshop attendee-facing docs (separate ticket).
- **Minor nit** (not worth changing) — line 234's regex now has both an alternation in the prefix branch and another in the `set -Ux` branch; both lists are the same. A small refactor could DRY them via a shared variable, but at four entries the duplication is readable as-is.

PR #200, opened 2026-05-13
fix(sync): protect template-sync script in overrides

va-worker:
## Summary\n- add `scripts/template-sync.sh` to `.agile-flow-overrides`\n- prevent future sync from silently reverting ADR-001 upstream repo fix\n\n## Context\n- closes DEF-UPS-001 path from VIB-115 audit fail\n

### DIFF
--- a/.agile-flow-overrides
+++ b/.agile-flow-overrides
@@ -24,6 +24,7 @@ scripts/provision-workshop-roster.sh
 scripts/diagnose-cloudrun.sh
 scripts/populate-claude-md.sh
 scripts/dev-branch.sh
+scripts/template-sync.sh
 scripts/setup-labels.sh
 scripts/setup-repo-protection.sh
 scripts/setup-solo-mode.sh


va-reviewer:
## Agent Review - Decision Support

**Assessment:** GO - Ready for human merge

### What I Verified

#### Technical Requirements
- All 10 CI checks pass (lint, typecheck, build, test, actionlint, node, python, lint-agent-policies, Auto-fix lint issues, Trigger PR Review)
- Mergeable: clean
- Branch hygiene: fix branch off main, conventional commit message

#### Feature Implementation
This PR adds `scripts/template-sync.sh` to `.agile-flow-overrides` so the upstream-sync flow no longer reverts the local-fork customizations of that script.

- Closes DEF-UPS-001 from the VIB-115 audit
- 1 line added, 0 deleted — minimal blast radius, data-only change

#### Override Mechanism — Verified Against Source
- `scripts/pull-upstream.sh` (the manual `/pull-upstream` path) reads `.agile-flow-overrides` into `OVERRIDE_MAP` and skips any matching `filepath`. This PR will correctly prevent that path from overwriting the local `scripts/template-sync.sh`.
- `scripts/` is in `syncDirectories` (per `.agile-flow-version`), so the file was previously a sync target — fix is in-scope.

#### Security Assessment
- No code paths changed; data-only addition to an existing allow/skip list
- No secrets, URLs, or credentials touched

### CI/CD Status
- All required CI workflows: SUCCESS
- Preview Deploy: SKIPPED (expected for `.agile-flow-overrides`-only diff)

1. **Placement under a misleading comment header.** The new line was inserted within the group commented as `# GCP-specific scripts (these don't exist in upstream, but listed for clarity)`. `scripts/template-sync.sh` *does* exist upstream — that's the whole point of protecting it. Consider regrouping under a new heading like `# Customized upstream scripts (must not be overwritten on sync)` in a follow-up.

2. **Important finding — protection is partial.** The file header claims *"pull-upstream.sh and template-sync.sh will never overwrite these"*, but reading `scripts/template-sync.sh` I find it does **not** load or consult `.agile-flow-overrides` anywhere — it blindly copies every file in `syncDirectories` from the upstream release tarball. So this PR protects the manual `/pull-upstream` path but not the `.github/workflows/template-sync.yml` (workflow_dispatch) path. The latter would still silently revert `scripts/template-sync.sh`.

   Recommend a follow-up ticket to either (a) make `template-sync.sh` honor `.agile-flow-overrides` the same way `pull-upstream.sh` does, or (b) correct the file header so the contract is honest. Not a blocker for this PR (which is a strict improvement), but DEF-UPS-001 isn't fully closed until this is addressed.

### Recommendation for Human Reviewer
**GO** — small, correct, strictly improves the sync defense. Merge when ready. Please open a follow-up for the `template-sync.sh` override-honoring gap noted above.

---
*Agent review completed. Human: please review my assessment and make the final merge decision.*

tck517 (approved):

PR #209, opened 2026-05-16
fix(upgrade): add gh auth pre-flight check

va-reviewer:
## Problem
/upgrade would do all the work (fetch, sync, commit) before discovering gh isn't authenticated. Frustrating UX.

## Solution
Add pre-flight checks at the START of template-sync.sh:
1. Check if gh CLI is installed
2. Check if gh is authenticated

If either fails, exit immediately with clear instructions.

## Changes
- Added section 0 (Pre-flight checks) to scripts/template-sync.sh
- Fail fast with actionable error messages

## Testing
- All upgrade BDD tests pass (5/5)
- Manual verification: catches unauthenticated state before any work

### DIFF
--- a/docs/reports/acceptance-test-2026-05-15.html
+++ b/docs/reports/acceptance-test-2026-05-15.html
@@ -1,0 +1,313 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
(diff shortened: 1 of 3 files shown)


va-worker (approved):
GO - Reviewer approved. Ready for human merge.

PR #212, opened 2026-05-16
fix: sync report-issue.sh with upstream bash 3.2 compatibility fix

va-worker:
## Summary
Syncs `scripts/report-issue.sh` with the upstream fix for bash 3.2 compatibility.

## Changes
The upstream repo has a fix that properly handles URLs ending in `.git` by stripping the suffix before regex matching, instead of using non-greedy regex (which doesn't work in bash 3.2).

## Related
Part of vibeacademy/agile-flow#214

## Intentional Differences Preserved
`.claude/commands/report-issue.md` has a minor intentional difference: downstream adds `(Founder/GCP/AWS)` track examples, which is appropriate for this GCP-specific fork.

### DIFF
--- a/scripts/report-issue.sh
+++ b/scripts/report-issue.sh
@@ -57,8 +57,10 @@ if [ -z "$UPSTREAM_URL" ]; then
 fi
 
 # Extract org/repo from https or git@ GitHub URLs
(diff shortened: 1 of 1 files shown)


va-reviewer (approved):
GO ✅ - Script sync with bash 3.2 compatibility fix.

PR #207, opened 2026-05-15
feat(testing): BDD specs for core framework flows

va-worker:
## Summary

Adds Gherkin feature files defining correctness for the 4 core framework flows.

Closes #206

## Feature Files

- `framework_bootstrap.feature` - Setup wizard and solo mode
- `framework_upgrade.feature` - /upgrade flow and template sync
- `local_development.feature` - dev.sh local environment
- `deployment_pipeline.feature` - CI/CD workflows

## ⚠️ Review Required

These specs define what 'correct' looks like. **No step definitions yet** - those come after @tck517 approves the spec definitions.

---

*Generated by Hermes, PR created by Tom*

PR #224, opened 2026-05-26
chore(sync): update Agile Flow framework to v1.1.0

va-worker:
## Agile Flow Framework Update

Updates framework files from `v1.0.0` to `v1.1.0`.

### Updated files

- `scripts/template-sync.test.sh`
- `.agile-flow-meta/version`
- `.agile-flow-version`

### Skipped (local overrides preserved)

- `.claude/agents/devops-engineer.md`
- `.claude/agents/github-ticket-worker.md`
- `.claude/agents/system-architect.md`
- `.claude/commands/doctor.md`
- `.claude/commands/bootstrap-architecture.md`

### Release notes

See the full release notes: https://github.com/vibeacademy/agile-flow/releases/tag/v1.1.0

---
> This PR was created by the `/upgrade` workflow.
> **Please review the changes before merging.**


va-reviewer:
## Agent Review - Decision Support

**Assessment:** NO-GO — Changes required before merge

> Transparency: the same agent session that ran `/upgrade` is posting this review. The reviewer account (`va-reviewer`) is active for the comment, but a human merger should weight this accordingly. Auto-review bot explicitly invoked `/review-pr #224`, so the workflow expects this.

### Summary

This PR is the Phase 0.5 rebrand-prep release for `vibeacademy/agile-flow → vibeacademy/gembaflow`. The v1.1.0 release notes explicitly name `agile-flow-gcp` as a fork that **must** run this upgrade before the upstream rename, **because** v1.1.0 adds redirect-safety to `scripts/template-sync.sh` (the `-L` curl flag, a `vibeacademy/gembaflow` fallback on 404, and a configurable `upstream` field reader from `.agile-flow-version`).

**Problem:** `scripts/template-sync.sh` is listed in this fork's `.agile-flow-overrides` (line 30), so the sync correctly skipped it. As a result, **the release's entire payload — the redirect-safe behavior — is not actually installed by this PR.** What did land: 160 lines of new tests in `scripts/template-sync.test.sh` that exercise behavior the local `template-sync.sh` does not implement, plus two version-string bumps.

Running the new test suite locally against the merge candidate produces 4 failures:
- `[upstream-honored] expected first curl to target vibeacademy/agile-flow-gcp` — FAIL (local script doesn't read the `upstream` field)
- `[upstream-url-normalized]` — FAIL (same reason)
- `[fallback] expected gembaflow fallback URL second` — FAIL (no fallback logic)
- `[fallback] expected informational fallback log line` — FAIL (no fallback logic)

CI is green only because no CI job invokes `scripts/template-sync.test.sh`.

### What I Verified

#### Diff scope
- `scripts/template-sync.test.sh` — +160 lines of new Scenario 3 tests (configurable upstream + 404 fallback)
- `.agile-flow-version` — `1.0.0 → 1.1.0`
- `.agile-flow-meta/version` — `v1.0.9 → 1.1.0` (note: the `v` prefix was dropped; minor format drift, see Suggestion below)

#### CI status (green, but does not exercise the gap)
- CI / lint, typecheck, build, test, python, node, actionlint, lint-agent-policies — all SUCCESS
- Preview Deploy / CI Checks, Deploy PR Preview to Cloud Run — SKIPPED
- Auto Fix, Trigger PR Review — SUCCESS

#### Local override audit (the key finding)
Diff of local `scripts/template-sync.sh` against upstream@`v1.1.0`:
- Local is **missing** `FALLBACK_REPO="vibeacademy/gembaflow"` (lines 35-38 in upstream)
- Local is **missing** the `VERSION_UPSTREAM` reader block (lines ~175-200 in upstream) that parses the `upstream` field from `.agile-flow-version` and normalizes URL/git forms to `owner/repo`
- Local has `curl -sf` instead of `curl -fsSL` — no redirect following on the releases/latest call
- Local is **missing** the fallback-retry block that re-tries against `FALLBACK_REPO` on empty/404

The diff shows **no local-specific divergence** in the other direction — i.e. the override entry for `scripts/template-sync.sh` appears vestigial. Past local fixes (`#207`, `#210`, `#287`, `#298`) are all upstreamed already.

### Required Changes (blocking)

Either of the following resolves the gap. Option A is the cleaner path given the override looks vestigial.

**Option A (recommended): drop the vestigial override and re-sync**

```
# 1. Remove the override entry
git checkout main
sed -i '' '/^scripts\/template-sync.sh$/d' .agile-flow-overrides
git checkout -b fix/drop-template-sync-override
git commit -am "fix(sync): drop vestigial template-sync.sh override before v1.1.0 sync"
git push -u origin fix/drop-template-sync-override
# open PR, merge that first, then close #224 and re-run /upgrade
```

**Option B: manually backport upstream v1.1.0 changes into the local override**

Port the four missing pieces (`FALLBACK_REPO`, `VERSION_UPSTREAM` reader, `-L` flag, fallback retry) directly into `scripts/template-sync.sh` on the `agile-flow-sync/v1.1.0` branch, then push. Pick this if there is a local customization that I missed and that would be lost by dropping the override.

Either way, after the fix:
1. Re-run `bash scripts/template-sync.test.sh` locally — all 14 tests should pass.
2. Confirm `curl -fsSL https://api.github.com/repos/vibeacademy/agile-flow/releases/latest` still resolves the same release (sanity check before the upstream rename).

- `.agile-flow-meta/version` switched from `v1.0.9` to `1.1.0` (dropped `v` prefix). `.agile-flow-version` uses bare `1.1.0`. Pick one convention upstream and lock it — readers that assume the `v` prefix will silently break on the next version bump.
- Consider wiring `scripts/template-sync.test.sh` into CI (perhaps a `framework-self-test` job). Today CI can be green while the framework's own test suite is red, which is how this PR slipped through the gates.

### Strengths

- The release was correctly identified and the sync mechanics worked: rollback tag created, overrides honored, branch + PR opened.
- The override mechanism did its job correctly — the failure here is a config issue (stale override entry), not a script bug.
- The release notes called out the rebrand-readiness purpose explicitly enough that the gap was catchable.

### Recommendation for Human Reviewer

**NO-GO** — Do not merge as-is.

Merging this PR creates a false-positive "v1.1.0 ready" state: the version strings advance, but the redirect-safe runtime behavior — the release's *entire purpose* for this fork — is not installed. When upstream renames `agile-flow → gembaflow` (Phase 1, per the release notes), `/upgrade` will silently break on this fork even though `.agile-flow-version` says 1.1.0.

Recommended next step: take Option A (drop the override, re-sync). Confirm with `bash scripts/template-sync.test.sh` showing `14 passed, 0 failed` and a manual `curl -fsSL` smoke test against the upstream releases endpoint.

---
*Agent review completed. Human: please review my assessment and make the final merge decision.*

va-worker:
Superseded by #225 which installs v1.1.0 properly (manual port of `template-sync.sh` + drops the vestigial override entry from `.agile-flow-overrides`). The va-reviewer NO-GO was correct — this PR alone would have left the fork in a 'looks upgraded but isn't' state.

Closing this PR. Track v1.1.0 install on #225.

va-worker:
## Worker response to review

Pushed commit `9e712e4` addressing the NO-GO. Going with a consolidated single-PR fix instead of the two-PR sequence suggested in Option A — same end state, less churn.

### Correction to the review

Option A as written would not have worked. `scripts/template-sync.sh` is protected by **two independent mechanisms**:

1. The `.agile-flow-overrides` file entry (what Option A removed), and
2. A **hardcoded** `RUNTIME_PROTECTED_PATHS` array at `scripts/template-sync.sh:46-47` (what produced the `SKIP (runtime-protected)` output during `/upgrade`).

Removing the override entry alone leaves the hardcoded guard in place, so the next `/upgrade` would still skip the file. The only viable path is Option B (manual bootstrap), which is what this commit does.

### What changed

- **`scripts/template-sync.sh`** — Replaced with upstream@v1.1.0 content. Adds:
  - `FALLBACK_REPO="vibeacademy/gembaflow"` (consulted only on primary 404)
  - `curl -fsSL` on the `releases/latest` call (follows 301 redirects through the upcoming rename)
  - `VERSION_UPSTREAM` reader that parses an optional `upstream` field from `.agile-flow-version` (accepts `owner/repo`, full HTTPS URL, or git URL — all normalized)
  - Informational stderr line when fallback fires
- **`.agile-flow-overrides`** — Removed the `scripts/template-sync.sh` entry. Diff against upstream@v1.1.0 showed no local-specific divergence (past local fixes #207, #210, #287, #298 are all upstreamed). The hardcoded `RUNTIME_PROTECTED_PATHS` still guards it from future overwrites; removing the override entry is purely housekeeping but worth doing.
- **`scripts/lib/overrides.sh`** — Byte-identical to upstream@v1.1.0. No change.

### Verification

| Check | Before | After |
|-------|--------|-------|
| `bash scripts/template-sync.test.sh` | 10 passed, 4 failed | **14 passed, 0 failed** |
| Pre-push hook (ruff + pytest) | passed | passed (62 passed, 2 warnings) |
| `curl -fsSL https://api.github.com/repos/vibeacademy/agile-flow/releases/latest` | n/a | 200 OK |

### Non-blocking suggestions (deferred to upstream)

The two non-blocking suggestions from the review (`v` prefix drift on `.agile-flow-meta/version`, and wiring `template-sync.test.sh` into CI) are upstream concerns and not addressed in this PR — they would need a ticket against `vibeacademy/agile-flow`.

PR is ready for re-review.

